PR #404, opened 2025-10-22
feat(trace2tree): move traverse_subtree_and_print to base class and add traverse_subtree_parents_only_and_print for JaxTraceToTree

olehtika:
-feat(trace2tree): move traverse_subtree_and_print to base class and add traverse_subtree_parents_only_and_print for JaxTraceToTree
-Refs: https://github.com/AMD-AGI/TraceLens/issues/403

### DIFF
--- a/TraceLens/Trace2Tree/trace_to_tree.py
+++ b/TraceLens/Trace2Tree/trace_to_tree.py
@@ -267,6 +267,129 @@ class BaseTraceToTree(ABC):
                     event[TraceLens.util.TraceEventUtils.TraceKeys.Args]["Python id"]
                 ] = event[TraceLens.util.TraceEventUtils.TraceKeys.UID]
 
+    def traverse_subtree_and_print(
+        self,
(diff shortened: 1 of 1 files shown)


ajassani (changes requested):
this PR breaks the torch tree traverse_subtree_and_print method 
can u leave the torch tree class untouched for now. I will work on the refactoring later

olehtika:
> this PR breaks the torch tree traverse_subtree_and_print method can u leave the torch tree class untouched for now. I will work on the refactoring later

Integration and inheritance from the base class is implemented in a separate PR 
https://github.com/AMD-AGI/TraceLens/pull/364/ 
https://github.com/AMD-AGI/TraceLens/blob/252-rename-current-pytorch-trace2tree-to-pytorchtrace2tree/TraceLens/Trace2Tree/trace_to_tree.py#L549
but you decided to reject it. With that PR PyTorch tree traverse_subtree_and_print works fine.

PR #450, opened 2025-12-17
supports Rocprofv3 profile data

amd-vivekag:
Until now, tracelens was not able to consume rocprof data. With this support, rocprof profile data can be analyzed through tracelens.

### DIFF
--- a/tests/test_rocprof_perf_report.py
+++ b/tests/test_rocprof_perf_report.py
@@ -40,18 +40,6 @@ class TestRocprofParser:
         assert isinstance(data["rocprofiler-sdk-tool"], list)
         assert len(data["rocprofiler-sdk-tool"]) > 0
 
-    def test_load_invalid_file(self):
-        """Test that invalid files raise an error"""
-        with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as f:
-            f.write('{"invalid": "data"}')
-            temp_file = f.name
-
-        try:
-            with pytest.raises(ValueError, match="Not a valid rocprofv3 file"):
-                RocprofParser.load_rocprof_data(temp_file)
-        finally:
-            os.unlink(temp_file)
-
     def test_extract_kernel_events(self, rocprof_file):
         """Test extracting kernel events"""
         data = RocprofParser.load_rocprof_data(rocprof_file)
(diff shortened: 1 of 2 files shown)


TraceLens/Reporting/generate_perf_report_rocprof.py
```diff
@@ -0,0 +1,272 @@
...
+            request_install("openpyxl")
+
+        with pd.ExcelWriter(output_xlsx_path, engine="openpyxl") as writer:
+            for sheet_name, df in dict_name2df.items():
+                df.to_excel(writer, sheet_name=sheet_name, index=False)
+                print(f"  - Sheet '{sheet_name}' ({len(df)} rows)")
+
+        print(f"\nSuccessfully written to {output_xlsx_path}")
+
+    return dict_name2df
```
oyazdanb: where we use this returned output.

TraceLens/Reporting/rocprof_analysis.py
```diff
@@ -0,0 +1,360 @@
...
+            logger.warning("No kernel events found in rocprof data")
+            return pd.DataFrame(columns=["type", "time ms", "percent"])
+
+        # Calculate total time from metadata
+        init_time = self.metadata.get("init_time", 0) / 1000.0  # ns to us
+        fini_time = self.metadata.get("fini_time", 0) / 1000.0  # ns to us
+        total_time_us = fini_time - init_time
+
+        # Calculate kernel execution time
+        kernel_time_us = sum(event["dur"] for event in self.kernel_events)
```
oyazdanb: this does not consider overlap between kernels and sums all the time together. We need to consider overlap between kernels and calculate gpu busy time.

tests/test_rocprof_perf_report.py
```diff
@@ -0,0 +1,271 @@
...
+        
+        # Check that we have expected row types
+        types = df['type'].tolist()
+        assert 'total_time' in types
+        assert 'kernel' in types
+        
+        # Percentages should sum to ~100% (excluding total_time row)
+        non_total = df[df['type'] != 'total_time']
+        total_pct = non_total['percent'].sum()
+        assert 99 < total_pct <= 100.1  # Allow for rounding
```
oyazdanb: why we have such restrict limit here?

amd-vivekag:
> @amd-vivekag I can merge this when the additional diffs in compare_perf_reports_pytorch.py and compare_perf_reports_pytorch.md‎ are reverted. Apart from that it looks good. Thanks for adding the test! Can you also edit the .github/workflows/regression-tests.yml file so the test runs on every PR like the other tests.
> 
> There is some duplicated logic with GPUEventAnalyzer but we can discuss later on how to reduce that.

Thanks @ajassani  for your quick response.
Since the compare script might need a separate attention, so let me revert that from this PR and create a separate PR. So that other code doesn't get blocked due to this. 
I'll also try to simplify the compare script in the new PR if right now, it is confusing in current form.

PR #451, opened 2025-12-19
feat(reporting): enable GPU-only trace support in perf report generation

ajassani:
- Add automatic GPU-only trace detection in TreePerfAnalyzer
- Skip CPU-dependent analysis (ops_summary, perf_metrics, collective analysis) for GPU-only traces
- Enable short kernel analysis for GPU-only traces (groups by Kernel name only)
- Add test case for GPU-only trace support
- Always generate kernel_summary and short_kernel_study in tests


Ahmedhasssan-aig (commented):
Looks good to me.

PR #465, opened 2026-01-20
Add category to op_summary, op_unique_args, unified_perf_summary

### LINKED ISSUE #353: Add op category to ops_summary and ops_all tables
Specifically, adding current classification label (GEMM, conv, etc.) using the same classification heuristics that are applied to create the aggregated table. Additionally, other aggregation methods are desired. E.g. nn.module information that can be used for classification later.

Basically the issue is that we have an aggregated table, but there's no way to tell without rerunning the analysis/reimplementing heuristics, what actually contributed to the aggregated table.

gabeweisz:
Add category to op_summary, op_unique_args, unified_perf_summary

Fixes #353


TraceLens/TreePerf/tree_perf.py
gabeweisz: This was linting

gabeweisz:
I'm not sure which table "ops_all" is and assumed that was ops_unique_args

amoskvic:
> What seems to be happening is that some of the rows in at least one table are getting split because before we were just aggregating on op name, whereas now we are splitting on op name and op category. It appears that we are getting rows with the same name but multiple categories. Will investigate further and update the reference results if necessary

Yes, this sounds quite likely. But just to clarify: this row splitting is actually desired behavior. Like if we have ops with the same name & shape get involved in different categories, it's good that we're splitting them as it'd give more granularity & flexibility during aggregation.

PR #470, opened 2026-01-21
Update lint rule to only run black on changed files

gabeweisz:
This prevents failures due to existing files that don't match the linting rule


gabeweisz:
This run correctly shows lint failing: https://github.com/AMD-AGI/TraceLens/actions/runs/21226480211/job/61074674831?pr=470

spandoesai (approved):
LGTM

PR #498, opened 2026-02-18
Improvement to TraceDiff's kernel grouping precision

unagiboi:
Improvement to TraceDiff's logic to create much finer kernel groupings.

1. Position based comparison: If two sequences of children are of the same length, they are automatically matched. If not, they go through the traditional wagner fischer algorithm and are aligned based on position and name. Since nodes may now be combined even if they have different names, in the report, `lowest_common_ancestor_name` will appear as `name` if `trace1_name` == `trace2_name`. Otherwise, it will be `trace1_name | trace2_name`.
2. If a child exists in both traces but only one of them is on a GPU path, previously, the child not on a GPU path would not be considered. Now children not on GPU paths are considered if they have a counterpart in the other trace that is on a GPU path, to appropriately match them.
3. Removed redundant tree traversal in `calculate_diff_boundaries` function. Does the initial tree difference calculations and merging all in one pass.
4. If wagner fischer fails to match two nodes, children one level below each node is observed.  If the sequences of children below two nodes are identical, then these nodes are matched.


unagiboi:
Regression tests will fail until #495 is merged with main

devalshahamd:
Looks good to me!

PR #579, opened 2026-04-03
Fix TraceDiff bugs

### LINKED ISSUE #569: [Bug] TraceDiff doesn't capture multi-threaded traces
For multi-threaded traces, TraceDiff only captures events in the first thread

### LINKED ISSUE #570: [Bug] TraceDiff fails to capture kernels for traces with shallow trees
`_get_top_level_root` gets the highest root node (most often python function) of a cpu op. The function squashes the tree from the top until a node is encountered that has multiple children. For shallow trees, if every node has only child, the entire subtree is removed.

unagiboi:
Previously, TraceDiff could only parse through one root node. This is problematic for some traces (training traces) that contain many root nodes. Now TraceDiff properly parses through every root node.

Additionally, TraceDiff performs an initial tree compression via truncating the top level nodes that only have one children. Ex:

```
node A
|____ node B
      |____ node C
             |____ node D
             |____ node E
```
becomes
```
node A
|____ node D
|____ node E
```

This is problematic if every node in the subtree has only children. In these cases the entire tree compresses to nothing. Now, the tree compression will stop when a CPU operation or kernel is encountered.
TraceDiff tests in private repo pass.
Closes #569
Closes #570

### DIFF
--- a/TraceLens/TraceDiff/trace_diff.py
+++ b/TraceLens/TraceDiff/trace_diff.py
@@ -228,17 +228,21 @@ class TraceDiff:
                 while True:
                     children = current.get("children", [])
                     if len(children) == 1:
-                        current = tree.get_UID2event(children[0])
+                        child = tree.get_UID2event(children[0])
+                        if (child.get("cat") or child.get("category")) == "cpu_op":
+                            break
+                        current = child
                     else:
-                        children = current.get("children", [])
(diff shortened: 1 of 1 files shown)


spandoesai:
LGTM, these are well needed generalization changes

spandoesai (approved):
LGTM

PR #582, opened 2026-04-03
Fix aten::convolution_backward appearing as ConvolutionBackward()

### LINKED ISSUE #566: [Bug] Fix unified perf summary showing ConvolutionBackward0 instead of aten::convolution_backward.

unagiboi:
There are three conditions with which a node is used for unified_perf_analysis
1. A node is a leaf CPU op
2. A node has parents who linked to a `fwd_event` with a perf model
3. A node itself is a CPU op

Once of these conditions is met, traversal of the subtree stops.

Because the traversal goes top down, there are instances where condition 2 is met and traversal stops even when there are descendants who meet condition 3.
Condition 3 should be preferred over condition 2. ConvolutionBackward() was appearing in perf report because it met condition 2. aten::convolution_backward has a perf model but wasn't being added to report since it's below ConvolutionBackward(). 
Now if condition 2 is met, it checks all descendants to see if there is an op with a perf model. If so, it continues traversing. Condition 2 is only met if no descendants have perf models.

Closes #566


gabeweisz (approved):

spandoesai (approved):
LGTM

PR #587, opened 2026-04-08
Checks if uid is not None

unagiboi:
If `uid` was 0, then the condition `if uid` would return False, even though uid existed. Changed all `if uid` conditions to `if uid is not None`.


spandoesai:
Thanks for the PR @unagiboi! One thing stood out: at line 232, the original code passes for both cpu_op or cuda_runtime, but the new condition only passes for cpu_op. Is this an intended change?

unagiboi:
> Thanks for the PR @unagiboi! One thing stood out: at line 232, the original code passes for both cpu_op or cuda_runtime, but the new condition only passes for cpu_op. Is this an intended change?

No it was not. Thanks for catching that

PR #600, opened 2026-04-21
[Tracelens-fixes] -- Edge cases small fixes

### LINKED ISSUE #596: [Tracelens-fixes] -- experimenting small fixes --External
Minor fixes
## Summary Table

# TraceLens — Bug Report (Partial Fix Status)

Notes:  ✅ FIXED ALREADY main - means this is already fixed in external Tracelens. 

Tracking 11 bugs identified in the internal repo against the public `main` branch.
**4 fixed · 7 not fixed**

| # | Bug | File | Priority | Status |
|---|-----|------|----------|--------|
| 1 | Boolean passed as dict key | [`util.py:558`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/util.py#L558) | High | ✅ FIXED |
| 2 | None <= 100 TypeError crash | [`trace_to_tree.py:136`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/Trace2Tree/trace_to_tree.py#L136) | High | ✅ FIXED ALREADY main |
| 3 | @property + @staticmethod stacked | [`gpu_event_analyser.py:75`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/TreePerf/gpu_event_analyser.py#L75) | High | ✅ FIXED (removed the function) |
| 4 | hasattr() on a dict | [`perf_model.py:49`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/perf_model.py#L49) | High | ✅ FIXED ALREADY main|
| 5 | kernel_names used before assignment | [`perf_model.py:38`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/perf_model.py#L38) | High | ✅ FIXED ALREADY main |
| 6 | Wrong kwargs in bytes_bwd() | [`perf_model.py:133`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/perf_model.py#L133) | High | ✅ FIXED |
| 7 | IndexError after empty-list warning | [`util.py:120`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/util.py#L120) | High | ✅ FIXED ALREADY main |
| 8 | None dereference on missing launch event | [`trace_fuse.py:111`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/TraceFusion/trace_fuse.py#L111) | High | ✅ FIXED |
| 9 | Shallow copy mutates original replay IR | [`event_replay.py:417`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/EventReplay/event_replay.py#L417) | High | Not fixed in here- DIFFERENT PR |
| 10 | default_categorizer signature mismatch | [`trace_to_tree.py:321`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/Trace2Tree/trace_to_tree.py#L321) | High | ✅ FIXED |
| 11 | math.log10(0) crash | [`trace_fuse.py:136`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/TraceFusion/trace_fuse.py#L136) | Medium | ✅ FIXED |

---

## 1. Boolean passed as dict key

**File:** [`TraceLens/util.py#L558`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/util.py#L558)
**Status:** ❌ NOT FIXED

The `==` comparison is evaluated before `.get()`, producing `False` as the key. `event.get(False)` always returns `None`, so metadata events are never detected and fall through misclassified — silently corrupting JAX trace analysis.

```python
# Buggy — evaluates to event.get(False)
if event.get(
    TraceEventUtils.TraceKeys.Phase == TraceEventUtils.TracePhases.Metadata
):

# Fix
if event.get(TraceEventUtils.TraceKeys.Phase) == TraceEventUtils.TracePhases.Metadata:
```

---

## 2. None <= 100 TypeError crash

**File:** [`TraceLens/Trace2Tree/trace_to_tree.py#L136`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/Trace2Tree/trace_to_tree.py#L136)
**Status:** ✅ FIXED

Previously `event.get("pid")` was compared directly with `<= 100`, raising `TypeError` in Python 3 when the key was missing. The pid is now safely assigned to a variable before any use, with no bare comparison against `None`.

```python
# Fixed — pid stored safely before use
pid = event.get(TraceLens.util.TraceEventUtils.TraceKeys.PID)
```

---

## 3. @property + @staticmethod stacked

**File:** [`TraceLens/TreePerf/gpu_event_analyser.py#L75`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/TreePerf/gpu_event_analyser.py#L75)
**Status:** ❌ NOT FIXED

Stacking `@property` over `@staticmethod` returns the raw descriptor object instead of executing the method. Any consumer of `all_event_keys` receives a descriptor, not an iterable — silently broken with no exception raised.

```python
# Buggy
@property
@staticmethod
def all_event_keys():
    return itertools.chain(...)

# Fix
@classmethod
@property
def all_event_keys(cls):
    return itertools.chain(cls.gpu_event_keys, cls.cpu_event_keys)
```

---

## 4. hasattr() on a dict

**File:** [`TraceLens/PerfModel/perf_model.py#L49`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/perf_model.py#L49)
**Status:** ✅ FIXED

Previously `hasattr(self.param_details, "B")` checked for an object attribute rather than a dict key, always returning `False`. Batch dimension `B` was therefore always overwritten to `1`, corrupting all FLOPS calculations. Now correctly uses `"B" not in self.param_details`.

```python
# Fixed
if "B" not in self.param_details:
    self.param_details["B"] = 1
```

---

## 5. kernel_names used before assignment

**File:** [`TraceLens/PerfModel/perf_model.py#L38`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/perf_model.py#L38)
**Status:** ✅ FIXED

Previously, if neither `"kernel_names"` nor `"kernel_details"` existed in the event, `kernel_names` was never assigned and the subsequent `for` loop raised `UnboundLocalError`. Now initialised to `[]` before the conditional block.

```python
# Fixed — initialised before if/elif
kernel_names = []
if "kernel_names" in event and len(event["kernel_names"]) > 0:
    kernel_names = event["kernel_names"]
elif "kernel_details" in event and len(event["kernel_details"]) > 0:
    kernel_names = [kernel["name"] for kernel in event["kernel_details"]]
for kernel_name in kernel_names:
```

---

## 6. Wrong kwargs in bytes_bwd()

**File:** [`TraceLens/PerfModel/perf_model.py#L133`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/perf_model.py#L133)
**Status:** ❌ NOT FIXED

`bytes_bwd()` calls `self.bytes_func(..., bytes_per_element=bytes_per_element)` but `bytes_func` (defined at line 102) expects `bpe_mat1, bpe_mat2, bpe_bias, bpe_output`. Raises `TypeError` on every GEMM backward byte calculation.

```python
# Buggy
bytes_input_grad = self.bytes_func(
    M=self.M, N=self.K, K=self.N, bias=False,
    bytes_per_element=bytes_per_element,  # no such parameter
)
# FYI: This fix works, cause:  for two reasons:

  # 1. bias=False in both calls — bpe_bias is multiplied by 0 inside bytes_func, so its value is irrelevant.                                                                                                
  # 2. Single uniform dtype — bytes_bwd is designed with one bytes_per_element for all tensors, which is exactly what all subclasses pass (super().bytes_bwd(self.bpe)). Setting bpe_mat1 = bpe_mat2 =
  bpe_output = bytes_per_element correctly reflects that.  

# Fix
bytes_input_grad = self.bytes_func(
    M=self.M, N=self.K, K=self.N, bias=False,
    bpe_mat1=bytes_per_element, bpe_mat2=bytes_per_element,
    bpe_bias=bytes_per_element, bpe_output=bytes_per_element,
)
```

---

## 7. IndexError after empty-list warning

**File:** [`TraceLens/util.py#L120`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/util.py#L120)
**Status:** ✅ FIXED

Previously the code emitted a warning when `hlo_filename` was empty but continued execution, immediately hitting `hlo_filename[0]` and raising `IndexError`. An early `return {}` is now present after the warning.

```python
# Fixed
elif len(hlo_filename) == 0:
    logger.warning(
        f"No matching hlo_filenames for module '{module_name}' in {dir_name}. "
        "HLO metadata will be unavailable."
    )
    return {}
```

---

## 8. None dereference on missing launch event

**File:** [`TraceLens/TraceFusion/trace_fuse.py#L111`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/TraceFusion/trace_fuse.py#L111)
**Status:** ❌ NOT FIXED

`next(..., None)` returns `None` when no matching launch event is found. The very next line subscripts `launch_event["args"]` without a `None` guard, raising `TypeError: 'NoneType' object is not subscriptable`.

```python
# Buggy
launch_event = next(
    (e for e in events if e.get("cat") in ["cuda_runtime", "cuda_driver"]
     and "launch" in e.get("name", "").lower()),
    None,
)
self.linking_key = "correlation" if "correlation" in launch_event["args"] else "External id"

# Fix
if launch_event is None:
    self.linking_key = "External id"
else:
    self.linking_key = "correlation" if "correlation" in launch_event["args"] else "External id"
```

---

## 9. Shallow copy mutates original replay IR

**File:** [`TraceLens/EventReplay/event_replay.py#L417`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/EventReplay/event_replay.py#L417)
**Status:** ❌ NOT FIXED

`list.copy()` is a shallow copy — inner dicts are shared references. Mutating them replaces `TensorCfg` objects in the original `event_replay_IR` with plain dicts, breaking all subsequent `replay()` calls in the same session.

```python
# Buggy
list_pos_args_copy, list_kwargs_copy = list_pos_args.copy(), list_kwargs.copy()
list_pos_args_copy[idx]["value"] = val["value"].__dict__  # mutates original IR

# Fix
import copy
list_pos_args_copy = copy.deepcopy(list_pos_args)
list_kwargs_copy = copy.deepcopy(list_kwargs)
list_pos_args_copy[idx]["value"] = val["value"].__dict__
```

---

## 10. default_categorizer signature mismatch

**File:** [`TraceLens/Trace2Tree/trace_to_tree.py#L321`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/Trace2Tree/trace_to_tree.py#L321)
**Status:** ❌ NOT FIXED

The base class calls `self.default_categorizer()` with no arguments, but `JaxTraceToTree` overrides it as a `@staticmethod` expecting one argument `event`. Raises `TypeError` on every `JaxTraceToTree` instantiation.

```python
# Buggy
@staticmethod
def default_categorizer(event: dict) -> str:
    ...

# Fix
def default_categorizer(self) -> None:
    self._categorizer = TraceEventUtils.prepare_event_categorizer
```

---

## 11. math.log10(0) crash

**File:** [`TraceLens/TraceFusion/trace_fuse.py#L136`](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/TraceFusion/trace_fuse.py#L136)
**Status:** ❌ NOT FIXED

When a field's max value is `0` (the `defaultdict` default for fields with no integer values), `math.log10(0)` raises `ValueError: math domain error`.

```python
# Buggy
field: 10 ** (math.ceil(math.log10(max_value)) + 1)

# Fix
field: 10 ** (math.ceil(math.log10(max(max_value, 1))) + 1)
```

janmatai:
# Pull Request Template

> **Note to AMDers:**  

**See the issues here: Fixes #596**

Testing the bugs -- after fix. 

  Code block (triple backticks):                                                                                                                                                                          
  ```bash                                                                                                                                                                                                 

# BUG 1 
python -m pytest tests/test_bug_verification.py::TestBug1BooleanDictKey -v -s   

# BUG 3 
python -m pytest tests/test_bug_verification.py::TestBug3PropertyStaticmethod -v -s

# BUG 6
python -m pytest tests/test_bug_verification.py::TestBug6BytesBwdWrongKwargs -v -s

# Run Test for BUG 8, 10, and 11
python -m pytest tests/test_bug_verification.py::TestBug8NullLaunchEvent tests/test_bug_verification.py::TestBug10DefaultCategorizerSig tests/test_bug_verification.py::TestBug11Log10Zero -v -s

# Or run test for all 
python -m pytest tests/test_bug_verification.py -v -s

                                                                                                                        
  ```


TraceLens/TraceFusion/trace_fuse.py
```diff
@@ -133,7 +138,7 @@ def _set_offset_multiplier(self, events):
                 if isinstance(value, int):
                     max_values[field] = max(max_values[field], value)
         self.offset_multiplier = {
-            field: 10 ** (math.ceil(math.log10(max_value)) + 1)
+            field: 10 ** (math.ceil(math.log10(max(max_value, 1))) + 1)
```
gabeweisz: I'm suspicious of this one - if this is a multiplier, are we sure we do not want to force it to 1 if the max value is 0? And I am not sure if we want to use a multiplier of 10 for a max value of 0.5 for example. But I don't have the full context for this bit of code. @ajassani what is your opinion?

janmatai: Good point. Based on my reading of the code:
                                                                                                                                                                                                          

1.  max_values won't get float values because of the isinstance(value, int) guard here:

  https://github.com/AMD-AGI/TraceLens/blob/3900438468c4a17b225dcfd9bdb541034d5de92d/TraceLens/TraceFusion/trace_fuse.py#L138 — only integers pass through.

2. For the max_value == 0 case, maybe cleaner to skip the field entirely rather than forcing it to 1:

                                                                                                                                                                                                          
`  self.offset_multiplier = {
      field: 10 ** (math.ceil(math.log10(max_value)) + 1)                                                                                                                                                 
      for field, max_value in max_values.items()                                                                                                                                                          
      if max_value > 0                                                                                                                                                                                    
  }   `

gabeweisz: That is fair - I recommend changing it to the cleanest fix possible if we are touching this code

janmatai: Sounds great! Thanks, @gabeweisz !  

@ajassani --- any suggestions here?

ajassani (commented):
**Bug 3 (`all_event_keys`):** This property isn't referenced anywhere outside its own definition â€” no callers in the codebase. Rather than fixing the `@property + @staticmethod` stacking (which `@classmethod + @property` does, but that combo was deprecated in 3.11 and removed in 3.13), could we just remove `all_event_keys` entirely? Dead code is the cleanest fix.

If we do want to keep it for future use, a plain class attribute is simpler and forward-compatible:
```python
all_event_keys = list(itertools.chain(gpu_event_keys, cpu_event_keys))
```

TraceLens/TreePerf/gpu_event_analyser.py
```diff
@@ -72,12 +72,10 @@ def subtract_intervalsA_from_B(intervals_to_subtract, intervals):
...
-    @staticmethod
-    def all_event_keys():
-        return itertools.chain(
-            GPUEventAnalyser.gpu_event_keys, GPUEventAnalyser.cpu_event_keys
-        )
+    def all_event_keys(cls):
+        return itertools.chain(cls.gpu_event_keys, cls.cpu_event_keys)
 
     def get_gpu_event_lists(self):
         """
```
ajassani, line 76: `all_event_keys` has no callers in the codebase - just remove it.

TraceLens/EventReplay/event_replay.py
```diff
@@ -414,7 +415,9 @@ def get_repro_info(self) -> Dict[str, Any]:
             self.event_replay_IR["list_kwargs"],
         )
         # Convert TensorCfg to dict for JSON serialization
-        list_pos_args_copy, list_kwargs_copy = list_pos_args.copy(), list_kwargs.copy()
+        list_pos_args_copy, list_kwargs_copy = copy.deepcopy(
+            list_pos_args
+        ), copy.deepcopy(list_kwargs)
         for idx, val in enumerate(list_pos_args_copy):
             if isinstance(val["value"], TensorCfg):
                 list_pos_args_copy[idx]["value"] = val["value"].__dict__
```
ajassani, line 420: This fix is also in the upcoming EventReplay PR - can we drop it from this one to keep the merge clean?

janmatai: @ajassani -- sounds god. I will remove them in that case.

TraceLens/TreePerf/gpu_event_analyser.py
janmatai: @ajassani  -- sounds good -- will do it.

TraceLens/TraceFusion/trace_fuse.py
ajassani: I think `math.log10(max_value + 1)` is cleaner here - handles zero without needing a separate filter or clamp.

ajassani (changes requested):

TraceLens/TraceFusion/trace_fuse.py
janmatai: @ajassani  -- this done. THe commit is here:  https://github.com/AMD-AGI/TraceLens/pull/600/commits/83d284d3ce173f36a30137c67a2596305cd3c0a7

TraceLens/TreePerf/gpu_event_analyser.py
janmatai: This is also done: https://github.com/AMD-AGI/TraceLens/pull/600/commits/484fec27fbcf5d4620d89d3b6d28209a187c3527

TraceLens/EventReplay/event_replay.py
janmatai: This is also done: https://github.com/AMD-AGI/TraceLens/pull/600/commits/4250176b4faf36a01966587c0139af42ea886183

ajassani (approved):

PR #603, opened 2026-04-23
fix(trace2tree): attach "bleeding" host events to a containing ancestor

kyle-hoffmeyer:
## Summary

`build_host_call_stack_tree` silently dropped any host event whose end timestamp extended past the current stack top, leaving the event with `parent = None` and omitting it from the tree. In real profiles this happens when a CUDA runtime
event (e.g. `hipLaunchKernel`) is timestamped slightly inside a sibling CPU op but ends after it. 

This fix pops the violating stack top so the event attaches to the nearest ancestor that actually contains it (Perfetto already renders the tree this way).

## Example

One `hipLaunchKernel` (E) starts 0.15 µs before `aten::empty_strided` (D) ends, but runs 7.78 µs — ending ~5 µs past D. The enclosing `aten::native_layer_norm_backward` (C) comfortably contains E.

```
time (µs)  →    0       10      20      30      40      50      60      70      80      90     100
                |───────|───────|───────|───────|───────|───────|───────|───────|───────|───────|
depth 0  (A)   ┌────────────────────────────────────────────────────────────────────────────────┐
               │  autograd::evaluate_function (A)                                               │
               │                  dur = 100                                                     │
               └────────────────────────────────────────────────────────────────────────────────┘
depth 1  (B)        ┌────────────────────────────────────────────────────────────────┐
                    │  NativeLayerNormBackward0 (B)                 dur = 80                    │
                    └────────────────────────────────────────────────────────────────┘
depth 2  (C)                ┌────────────────────────────────────────────────┐
                            │  aten::native_layer_norm_backward(C)         dur = 60      │
                            └────────────────────────────────────────────────┘
depth 3  (D)                                                ┌───┐
                                                            │ D │  aten::empty_strided
                                                            └───┘    dur = 5
depth 3  (E)                                                   ┌────────┐          ← BLEEDS past D
                                                               │   E    │  hipLaunchKernel
                                                               └────────┘    dur = 10
                                                                  ▲
                                                                  │  ac2g flow
                                                                  ▼
GPU stream                                                     ┌──────┐              (much later)
                                                               │  K   │  gpu_kernel
                                                               └──────┘
(intervals: A=[0,100]   B=[5,85]   C=[15,75]   D=[50,55]   E=[54,64])
```

### Before

When processing E, the working stack is `[A, B, C, D]`.

- Old pop condition only pops if `E.ts >= top.end` → does not pop D (`E.ts 53.8 < D.end 53.9`).
- A secondary check then sees `E.end 61.6 > D.end 53.9` and executes `continue`, so E is never given a parent and never pushed onto the stack.

Result: `E.parent = None`. Kernel K's parent chain is `K → E → None`, so it is treated as an orphan — no `cpu_op` ancestor, rendered as a `Synthetic Op` in `unified_perf_summary`.

### After

The pop loop is widened to also pop any stack frame that cannot fully contain the incoming event (`event.end > top.end`). On E:

1. Pop D (E.end > D.end).
2. New top is C; C fully contains E → stop, set `E.parent = C`, push E.

Result: `E.parent = C`, and K's parent chain is now `K → E → C (aten::native_layer_norm_backward) → B → A`. The kernel no longer produces a `Synthetic Op` row.


unagiboi:
@ajassani, the change in reference reports were due to there being 74 events not being captured in first trace and 73 in second trace. These events are now captured with the changes in this branch. These 73/74 extra events led to the UID of the events being off by a fixed number. You can verify this by subtracting the test `ex_UID` and reference `ex_UID` number for each mismatch in the failing regression test.
These missed events however were python functions that did not call any kernels. Therefore, they did not lead to any new rows being added or any kernels being added to existing rows, hence why only `ex_UID` is the differing column.

I've updated the reference reports.

ajassani (approved):

PR #607, opened 2026-04-28
EventReplay: extend beyond aten ops with auto-import and custom initializers

ajassani:
## Summary

EventReplay previously only worked with `aten::` ops. This PR extends it to support **custom ops from any namespace** (vLLM, aiter, etc.) and adds **custom initializers** so data-dependent ops produce realistic replay behavior.

### Extending beyond aten

- **Auto-import for custom op namespaces**: when EventReplay encounters a non-aten op (e.g., `_rocm_C::paged_attention`, `aiter::ck_moe_stage1`), it automatically imports the library that registers the op's schema. Supports `aiter`, `_rocm_C`, `_C`, `vllm` out of the box, and users can register additional namespaces.
- **Schemaless fallback**: ops that lack a registered schema can still be replayed with heuristic type inference from the profiler data.

### Custom initializers for data-dependent ops

- **`PagedAttentionInit`** — fills `block_tables`, `seq_lens`, and `query_start_loc` with realistic values so the attention kernel does real work instead of short-circuiting on zeros.
- **`MoeRoutingInit`** — constructs a complete token-to-expert routing table (`sorted_token_ids`, `sorted_expert_ids`, `num_valid_ids`) with configurable distribution (uniform or Zipf).
- **User-extensible**: subclass `CustomInit`, set `op_patterns` to the exact op name, implement `initialize()`, and register. First-match-wins, exact name matching.

### Iteration annotations (vLLM)

- `extract_batch_context` parses vLLM's per-iteration `user_annotation` events to get the exact prefill/decode split, so `PagedAttentionInit` builds an accurate `query_start_loc` for mixed batches.

### Bug fixes

- `replay()` with `lazy=True` + `auto_init=True` crashed (`AttributeError`) — fixed
- `get_repro_info()` corrupted `event_replay_IR` via shallow copy — fixed
- `batched_replay.py`: `benchmark_func` dict return type crash + dead `--op-filter`/`--op-limit` flags — fixed
- `replay()` now returns the op result
- Custom init matching changed from substring to exact name

### Tests and docs

- 11 CPU-only unit tests + GPU integration test with kernel name validation
- Docs rewritten: IR interpretability examples, step-by-step custom init guide, iteration annotations explained

## Test plan

- [x] CPU unit tests pass (11/11, ~3s)
- [x] GPU integration tests pass (5/5 ops, MI300X)
  - Kernel name match: all MATCH
  - Lazy mode, get_repro_info idempotency, return value, first-match-wins: all PASS


TraceLens/EventReplay/event_replay.py
```diff
@@ -63,35 +305,43 @@ def _setup(self):
...
+            self.args, self.kwargs = EventReplayer._get_args_kwargs(
                 self.event_replay_IR, device=self.device
             )
-        else:
-            args, kwargs = self.args, self.kwargs
 
-        # Call the function with the arguments
-        func(*args, **kwargs)
+        if not self._inits_applied and self._auto_init:
+            self._apply_custom_inits()
```
ajassani, line 318: **Blocking — lazy + custom init is wrong on the 2nd `replay()`.**

Lazy rebuilds `self.args` from the IR every call, but `_inits_applied` stays `True`, so paged-attn / MoE tensors go back to zeros after the first iteration.

Re-apply inits whenever lazy rebuilds args (or reset `_inits_applied` when args are reconstructed).

TraceLens/EventReplay/event_replay.py
```diff
@@ -19,28 +20,242 @@
...
 class EventReplayer:
+    _custom_init_registry: List[CustomInit] = [
+        PagedAttentionInit(),
+        MoeRoutingInit(),
+    ]
+
+    @classmethod
+    def register_custom_init(cls, init: CustomInit):
+        """Add a custom initializer to the registry."""
+        cls._custom_init_registry.append(init)
```
ajassani, line 216: **Blocking — user inits can never override builtins.**

`register_custom_init` appends, and `_apply_custom_inits` is first-match-wins. A user init for `_rocm_C::paged_attention` will never run.

Insert at the front, or skip the builtin when a later init matches the same exact name.

TraceLens/EventReplay/event_replay.py
```diff
@@ -19,28 +20,242 @@
...
+    """Try to import the library that registers a custom op's schema.
+
+    Returns True if at least one new module was successfully imported.
+    """
+    namespace = op_name.split("::")[0] if "::" in op_name else ""
+    if not namespace or namespace == "aten":
+        return False
+    if namespace in _auto_import_attempted:
+        return False
+    _auto_import_attempted.add(namespace)
```
ajassani, line 59: **Sticky even on failed import.** This set is process-global and is populated before the import succeeds. After a miss, `register_namespace()` never retries.

Only record successful attempts, or drop the namespace from the set inside `register_namespace`.

TraceLens/EventReplay/event_replay.py
```diff
@@ -49,10 +264,37 @@ def _setup(self):
...
+            )
+            self._schemaless = False
+        except ValueError:
+            if self.verbose:
+                print(
+                    "No schema found; falling back to schemaless replay "
+                    "(all args treated as positional, types inferred from profile)"
+                )
+            self.matched_schema = None
+            self._schemaless = True
```
ajassani, line 288: Schemaless fallback swallows the `ValueError` and proceeds with heuristic types. Unless `verbose=True`, a wrong-signature custom-op call looks like success.

Please `logger.warning` here (op name + how many schemas were searched).

TraceLens/EventReplay/event_replay.py
```diff
@@ -267,36 +507,60 @@ def _get_event_replay_IR(
...
-                        # as it will be filled with a value later
-                        init = None
-                    elif evt_name == "aten::copy_" and arg_name != "src":
-                        # special case for copy_ where we don't need to initialize the tensor
-                        # as it will be copied from another tensor
+                    if EventReplayer._should_skip_tensor_init(evt_name, arg_name, idx):
                         init = None
+                    profiled_dtype = event["args"]["Input type"][idx]
+                    if profiled_dtype in ("long", "long int", "int", "bool", "unsigned char"):
+                        init = "zeros" if init == "normal" else init
```
ajassani, line 534: `char` / `short` were added to `list_profile_tensor_types` but not this zeros-init branch. `build_tensor(..., init="normal")` then raises on those integer dtypes.

Same gap in `_get_event_replay_IR_schemaless` (~631).

ajassani (commented):
**Review of #607 (inline comments on the specific lines).**

Not merge-ready yet: rebase onto current `main` first (this branch is ~202 commits behind; docs now live at `docs/how-to/event-replay.md`, CI is `pytest tests/ -m "not gpu" --cov-fail-under=95`, and `tests/test_event_replay.py` already exists). The custom-op resolution + custom-init design is the right shape; the inlined notes are the correctness/CI gaps.

TraceLens/EventReplay/event_replay.py
```diff
@@ -19,28 +20,242 @@
...
+    if namespace in _auto_import_attempted:
+        return False
+    _auto_import_attempted.add(namespace)
+
+    modules = _NAMESPACE_IMPORTS.get(namespace, [namespace])
+    imported_any = False
+    for mod in modules:
+        try:
+            __import__(mod)
+            print(f"[EventReplayer] Auto-imported '{mod}' for op '{op_name}'")
```
ajassani, line 66: Unconditional `print` on every auto-import. Prefer `logger.info` (and keep the miss path behind `verbose`).

TraceLens/EventReplay/event_replay.py
```diff
@@ -19,28 +20,242 @@
...
+}
+
+# -- Auto-import registry --------------------------------------------------
+_NAMESPACE_IMPORTS: Dict[str, List[str]] = {
+    "aiter": ["aiter"],
+    "_rocm_C": ["vllm._rocm_C"],
+    "_C": ["vllm._C"],
+    "_C_cache_ops": ["vllm._C"],
+    "vllm": ["vllm._C", "vllm._rocm_C"],
+}
```
ajassani, line 44: Bare `_C` → `vllm._C` will mis-import if another library owns `_C`. Worth documenting the collision, and maybe not binding bare `_C` by default.

TraceLens/EventReplay/batched_replay.py
```diff
@@ -99,11 +99,17 @@ def _get_args_kwargs_from_ir(
     replayed_count = 0
     errors = 0
 
-    for i, repro_info in enumerate(repro_data_list):
+    ops_to_replay = repro_data_list
+    if args.op_filter:
+        ops_to_replay = [r for r in ops_to_replay if args.op_filter in r["op_name"]]
+    if args.op_limit:
+        ops_to_replay = ops_to_replay[: args.op_limit]
```
ajassani, line 106: **Blocking — CLI still cannot replay custom ops.** `--op-filter` / `--op-limit` are wired up, but resolution below is still `torch._C._jit_get_operation` only (line 116). This CLI cannot replay the `aiter` / `_rocm_C` / `_C` ops this PR exists for.

Use `_resolve_op_func` (or the same JIT → torch.ops → auto-import path).

TraceLens/EventReplay/custom_inits.py
```diff
@@ -0,0 +1,390 @@
...
+realistic content so the GPU kernel exercises real memory-access and compute
+patterns during replay benchmarking.
+
+To add a custom initializer for a new op family:
+  1. Subclass ``CustomInit``
+  2. Set ``op_patterns`` to one or more substrings that match the op name
+  3. Implement ``initialize()`` — mutate replayer.args / replayer.kwargs in-place
+  4. Return a one-line summary string (printed by EventReplayer)
+  5. Register with ``EventReplayer.register_custom_init(YourInit())``
+     or add it to the ``_custom_init_registry`` default list.
```
ajassani, line 21: Stale vs the implementation: matching is exact name (`op_name in self.op_patterns`), not substrings. And `register_custom_init` currently appends after builtins, so step 5 cannot override `PagedAttentionInit` / `MoeRoutingInit`.

TraceLens/EventReplay/custom_inits.py
```diff
@@ -0,0 +1,390 @@
...
+
+    if not annotations:
+        return 0
+
+    annotations.sort(key=lambda a: a["ts"])
+
+    annotated = 0
+    for e in analyzer.tree.events:
+        name = e.get("name", "")
+        if "paged_attention" not in name:
```
ajassani, line 94: This still uses substring (`"paged_attention" in name`) while `CustomInit.applies_to` is exact-name. Easy to annotate a similarly named op that the init will then ignore — or miss a renamed exact op.

TraceLens/EventReplay/custom_inits.py
```diff
@@ -0,0 +1,390 @@
...
+        def _by_name_or_pos(name, pos):
+            if name in arg_names:
+                return args[arg_names.index(name)]
+            return args[pos]
+
+        block_tables = _by_name_or_pos("block_tables", 9)
+        seq_lens = _by_name_or_pos("seq_lens", 10)
+        key_cache = _by_name_or_pos("key_cache", 5)
+        block_size = int(_by_name_or_pos("block_size", 12))
+        max_seq_len = int(_by_name_or_pos("max_seq_len", 13))
```
ajassani, line 164: Hardcoded positional fallback will silently init the wrong tensor if the schema renames/reorders args. Same pattern in `MoeRoutingInit` (~298).

Fail if the name is missing instead of guessing a slot.

TraceLens/EventReplay/test_event_replay.py
```diff
@@ -0,0 +1,204 @@
...
+# See LICENSE for license information.
+###############################################################################
+
+"""
+Tests for EventReplay core functionality.
+
+All tests use CPU-only ops (aten::mm) so they run without a GPU.
+Run from the repo root:
+    python -m pytest TraceLens/EventReplay/test_event_replay.py -v
+"""
```
ajassani, line 13: **Blocking for CI.** Current unit workflow is `pytest tests/ -m "not gpu" --cov-fail-under=95`. These 11 tests are not collected, and the new `custom_inits.py` / expanded `event_replay.py` will drop coverage.

Move this file under `tests/`. Also: `tests/test_event_replay.py` already exists on current `main` (GPU ResNet integration) — rebase will collide.

TraceLens/EventReplay/test_event_replay_gpu.py
```diff
@@ -0,0 +1,261 @@
...
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
+
+import torch
+from torch.profiler import profile, ProfilerActivity
+
+from TraceLens.EventReplay.event_replay import EventReplayer
+from TraceLens.EventReplay.custom_inits import CustomInit
+from TraceLens.EventReplay.utils import TensorCfg
+
+assert torch.cuda.is_available(), "GPU required for this test"
```
ajassani, line 31: Module-level `assert` means this file cannot be collected as a pytest test. Fine as a manual script — please don't name it `test_*.py` under a collected path after the rebase (or mark/skip at collection time).

TraceLens/EventReplay/test_event_replay_gpu.py
```diff
@@ -0,0 +1,261 @@
...
+        for e in all_events:
+            if e.get("cat") == "kernel" and e.get("name", ""):
+                orig_gpu.add(e["name"])
+
+        replay_gpu = set()
+        for e in replay_trace.get("traceEvents", []):
+            if e.get("cat") == "kernel" and e.get("name", ""):
+                replay_gpu.add(e["name"])
+
+        kernel_match = "MATCH" if replay_gpu.issubset(orig_gpu) else "MISMATCH"
```
ajassani, line 198: `orig_gpu` is **every** kernel in a 5-op trace. `replay_gpu.issubset(orig_gpu)` is almost always true even if this op launched the wrong kernel.

Compare against kernels from this cpu_op (or this op's original vs replayed kernel set).

ajassani:
@gabeweisz @kyle-hoffmeyer this is ready for another look. Rebased onto current main, and the blocking review items (docs/CI layout, lazy custom inits, CLI op resolve, user init override) are addressed.

PR #610, opened 2026-04-29
fix(treeperf): handle zero-duration GPU events in sweep-line

gphuang:
## Summary

Closes AMD-AGI/TraceLens-internal#356
Sub-issue of AMD-AGI/TraceLens-internal#182

`GPUEventAnalyser.get_gpu_event_lists()` raises `KeyError` when a PyTorch profile trace contains any GPU event with `dur == 0`. Fix is to skip zero-measure intervals from the sweep-line that computes `overlapping_uids` (they cannot overlap anything by construction). All 4 GPU event categories (`kernel`, `gpu_memcpy`, `gpu_memset`) are still included in the returned lists; only the overlap computation skips them.

### Bug

`get_gpu_event_lists()` builds two sweep-line points per GPU event and sorts by `(timestamp, point_type)`:

```python
points.append((event["ts"],   1, event["UID"]))   # start
points.append((event["t_end"], 0, event["UID"]))  # end
points.sort(key=lambda x: (x[0], x[1]))           # end (0) < start (1) at ties
```

The end-before-start tie-break is intentional: it prevents two events touching at a boundary from being flagged as overlapping. But for a single event with `dur == 0`, `t_end == ts`, so its **own** end-point is processed before its start-point and `active_uids.remove(uid)` raises:

```
File ".../TraceLens/TreePerf/gpu_event_analyser.py", line 148, in get_gpu_event_lists
    active_uids.remove(uid)
KeyError: 1821686
```

The reporter's failing trace (newer profiler image, Primus 26.2) contains exactly this shape; quick `jq` probe over the raw trace:

| Trace | Zero-duration GPU events |
|---|---|
| Working baseline | 0 |
| Failing trace | 2 × `gpu_memcpy` HtoD with `dur=0` |

The two events are 4 MB HtoD transfers reported with `dur=0` µs (visible in the trace's own `args.memory bandwidth (GB/s)` field as ~`2.27e-13`). The same algorithm trips on any zero-dur `kernel` / `gpu_memset` too — the bug is generic, just newly surfaced.

### Fix

`TraceLens/TreePerf/gpu_event_analyser.py`:

- Always initialise `event["overlapping_uids"] = set()` (so downstream readers do not NPE on skipped events).
- Only append `(start, end)` points to the sweep-line when `t_end > ts`. A zero-measure interval cannot overlap anything, so excluding it preserves the semantics of `overlapping_uids` exactly.
- Same caching behaviour preserved: a second call on a tree where `overlapping_uids` is already cached (zero-dur or not) is still a no-op.

Comment in the patched block names AMD-AGI/TraceLens-internal#182 so future readers can find the original report.

## Test plan

- [x] 4 new regression tests in `tests/test_gpu_event_overlapping_uids.py`:
    - `test_zero_duration_memcpy_does_not_raise` — direct repro shape (kernel + zero-dur memcpy + kernel)
    - `test_zero_duration_kernel_does_not_raise` — `cat=kernel` variant (future-proofs against sub-µs Triton / HIP-graph kernels)
    - `test_zero_duration_does_not_break_other_overlaps` — surrounding overlapping events still detected
    - `test_only_zero_duration_events_does_not_raise` — degenerate edge case
- [x] All 37 tests in `tests/test_gpu_event_overlapping_uids.py` pass (33 pre-existing + 4 new)
- [x] CI-relevant local test sweep passes: 123 tests across `test_subtract_intervals`, `test_graph_mode`, `test_kernel_launchers`, `test_perf_report_regression`, `test_compare_perf_reports`, `test_inference_perf_report`, `test_collective_analysis`, `test_all2allv_summary`, `test_detect_recompute`, `test_copyright_headers`, `test_gpu_event_overlapping_uids`
- [x] `black==26.1.0 --check` clean on both edited files
- [x] End-to-end repro on the original failing trace via `TraceLens_generate_perf_report_pytorch --extension_file examples/example_megatron_extension.py`: previously raised `KeyError: 1821686`; now writes the perf-report `.xlsx` successfully (~258 KB output, same shape as the working baseline trace's report)
- [x] Sanity re-run on a known-good trace (no regression): same output size pre/post-fix

### DIFF
--- a/TraceLens/TreePerf/gpu_event_analyser.py
+++ b/TraceLens/TreePerf/gpu_event_analyser.py
@@ -101,12 +101,21 @@ class GPUEventAnalyser:
                 if "t_end" not in event:
                     event["t_end"] = event["ts"] + event["dur"]
                 if "overlapping_uids" not in event:
-                    compute_overlapping_uids = True
-                    points.append(
-                        (event["ts"], 1, event["UID"])
-                    )  # 1 for start, 0 for end (end sorts first so boundary-touching != overlapping)
-                    points.append((event["t_end"], 0, event["UID"]))
                     event["overlapping_uids"] = set()
+                    # Skip zero-duration GPU events from the sweep-line below.
+                    # A zero-measure interval cannot overlap anything, and the
+                    # (ts, point_type) sort ordering puts end (0) before start
+                    # (1) at equal timestamps. For a zero-dur event, ts == t_end,
+                    # so its own end-point would be processed before its
+                    # start-point and raise KeyError on active_uids.remove(uid).
+                    # Observed on ROCm 7.2 traces (gpu_memcpy events with
+                    # dur=0); see TraceLens-internal#182.
+                    if event["t_end"] > event["ts"]:
+                        compute_overlapping_uids = True
+                        points.append(
+                            (event["ts"], 1, event["UID"])
+                        )  # 1 for start, 0 for end (end sorts first so boundary-touching != overlapping)
+                        points.append((event["t_end"], 0, event["UID"]))
                 gpu_events.append(event)
 
                 if category == "gpu_memcpy":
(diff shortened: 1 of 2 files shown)


gabeweisz (approved):
LGTM - small change that will prevent exceptions

PR #611, opened 2026-04-29
fix(treeperf): preserve kernel time when perf model fails

kyle-hoffmeyer:
When an operation is being added to `unified_perf_summary`, kernel metrics can be left blank or excluded from `unified_perf_summary` if:
- Operation name suggests there is a perf model, but perf model doesn't actually exist for the operation
- Operation parameters are missing or misconfigured

Kernel metrics (runtime) should not be dropped just because perf metrics (bandwidth, AI) could not be populated. Now, kernel metrics are populated regardless of if deriving perf metrics for it fails


TraceLens/TreePerf/tree_perf.py
```diff
@@ -2041,28 +2073,34 @@ def list_to_tuple(obj):
                         if k.startswith("param: ")
                     }
                     row["perf_params"] = perf_params if perf_params else None
-                except NotImplementedError:
-                    # This means we don't have a perf model for this op, which is expected for some ops. Ignore this.
-                    continue
                 except Exception as e:
                     perf_metrics_failed.append((event, e))
                     row["perf_params"] = None
+                    gpu_event_uids = event.get("gpu_events", [])
```
devalshahamd: Can we use the partial_metrics returned from  compute_perf_metrics instead of recalculating Kernel Time (µs) here?

gabeweisz (approved):

PR #615, opened 2026-04-30
Added option to include callstack in unified perf report; minor bug fix

devalshahamd:
### Add call stack support to unified perf report

Adds an `--include_call_stack` flag to both reporters that annotates `unified_perf_summary` with call stack information for downstream analysis. The key design choice is that call stacks are computed **post-aggregation** in `summarize_df_unified_perf_table` rather than per-event — this ensures call stack differences never influence grouping.

**Design**

Call stacks are computed after aggregation using `ex_UID` to look up one representative event per row. This avoids ops with trivially different stacks (e.g. the same op dispatched from different TorchInductor compiled graph files or through different paths in `compiler_interface.py`) splitting into separate rows despite being the same logical operation.

**Changes**

`tree_perf.py` — `summarize_df_unified_perf_table`:
- Added `include_call_stack=False` and `tree=None` parameters.
- After aggregation, uses `ex_UID` to look up one representative event per row and fetch its call stack via `tree.traverse_parents_and_get_callstack()`. Emits a warning if `include_call_stack=True` but `tree=None`.

Both reporters (`generate_perf_report_pytorch.py`, `generate_perf_report_pytorch_inference.py`):
- Added `include_call_stack: bool = False` function parameter and `--include_call_stack` CLI flag.
- Pass `include_call_stack` and `tree=perf_analyzer.tree` to the main `summarize_df_unified_perf_table` call.
- The overlap summary variant (`include_overlapping_kernels=True`) deliberately omits call stacks — it is used for concurrent kernel analysis, not call site attribution.
- When `include_call_stack=True`: writes `unified_perf_callstacks` (full call stacks, one row per summary row) as a separate sheet/CSV, and inserts `trunc_call_stack` (first 4 frames) into `unified_perf_summary`.

Example output:
When --include_call_stack used,

- A new column trunc_call_stack is included in the unified_perf_summary.csv (4 levels up)
- 
image

- A new sheet unified_perf_callstack is added with full callstack, correlated with unified_perf_summary though row_id

### DIFF
--- a/TraceLens/Reporting/generate_perf_report_pytorch.py
+++ b/TraceLens/Reporting/generate_perf_report_pytorch.py
@@ -256,6 +256,7 @@ def generate_perf_report_pytorch(
     enable_origami: bool = False,
     # activation recompute detection
     detect_recompute: bool = False,
+    include_call_stack: bool = False,
 ) -> Dict[str, pd.DataFrame]:
     if gpu_arch_json_path:
         with open(gpu_arch_json_path, "r") as f:
@@ -605,6 +606,8 @@ def generate_perf_report_pytorch(
                 agg_metrics=agg_metrics,
                 include_pct=True,
                 group_by_num_kernels=group_by_num_kernels,
+                include_call_stack=include_call_stack,
+                tree=perf_analyzer.tree,
             )
             if not df_unified_perf_summary.empty:
                 df_unified_perf_summary = add_truncated_kernel_details(
@@ -612,6 +615,25 @@ def generate_perf_report_pytorch(
                     source_col="kernel_details_summary",
                     new_col_name="trunc_kernel_details",
                 )
+                if "call_stack" in df_unified_perf_summary.columns:
+                    df_callstacks = df_unified_perf_summary[
+                        ["name", "op category", "call_stack"]
+                    ].copy()
+                    df_callstacks.insert(0, "row_id", range(len(df_callstacks)))
+                    dict_name2df["unified_perf_callstacks"] = df_callstacks
+
+                    n_frames = 4  # op name + 3 parent frames
+                    cs_col = df_unified_perf_summary.columns.get_loc("call_stack")
+                    df_unified_perf_summary.insert(
+                        cs_col,
+                        "trunc_call_stack",
+                        df_unified_perf_summary["call_stack"].apply(
+                            lambda s: " => ".join(str(s).split(" => ")[:n_frames])
+                        ),
+                    )
+                    df_unified_perf_summary = df_unified_perf_summary.drop(
+                        columns=["call_stack"]
+                    )
                 dict_name2df["unified_perf_summary"] = df_unified_perf_summary
 
             if include_overlap_info:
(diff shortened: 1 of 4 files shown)


TraceLens/Reporting/generate_perf_report_pytorch.py
```diff
@@ -256,19 +256,21 @@ def generate_perf_report_pytorch(
...
     detect_recompute: bool = False,
+    include_call_stack: bool = False,
 ) -> Dict[str, pd.DataFrame]:
     if gpu_arch_json_path:
         with open(gpu_arch_json_path, "r") as f:
             gpu_arch_json = json.load(f)
     else:
         gpu_arch_json = None
-
+    add_python_func = True if include_call_stack else False
```
kyle-hoffmeyer, line 266: This line isn't in generate_perf_report_pytorch_inference. As a result, python functions don't appear in call stack when generate_perf_report_pytorch_inference is used

TraceLens/Reporting/generate_perf_report_pytorch_inference.py
```diff
@@ -888,13 +889,34 @@ def generate_perf_report_pytorch(
...
                 df_unified_perf_summary = add_truncated_kernel_details(
                     df_unified_perf_summary,
                     source_col="kernel_details_summary",
                     new_col_name="trunc_kernel_details",
                 )
+                if "call_stack" in df_unified_perf_summary.columns:
+                    df_callstacks = df_unified_perf_summary[
+                        ["name", "op category", "call_stack"]
+                    ].copy()
+                    df_callstacks.insert(0, "row_id", range(len(df_callstacks)))
```
kyle-hoffmeyer, line 907: Do you think "row_id" is needed, if it's always going to be [0, n] and aligns with unified_perf_summary?

TraceLens/TreePerf/tree_perf.py
```diff
@@ -2292,6 +2296,24 @@ def summarize_df_unified_perf_table(
...
+
+            def _get_call_stack(ex_uid):
+                try:
+                    event = tree.get_UID2event(int(ex_uid))
+                    cs = tree.traverse_parents_and_get_callstack(
+                        event, filter=["nn.Module", "::", "/"]
+                    )
+                    return re.sub(r"_\d+", "", cs)
+                except Exception:
+                    return ""
```
kyle-hoffmeyer: I wonder if when there's an error it should return "ERROR" instead of empty string. That way, if it's blank, it won't look like the event is at the top of the tree

TraceLens/Reporting/generate_perf_report_pytorch.py
devalshahamd: Good point, let me add that. The inference file uses gorup_by_parent_module as well and add_python_func was enabled through that. I should make it consistent.

ajassani (approved):

TraceLens/TreePerf/tree_perf.py
devalshahamd: I added Not found

TraceLens/Reporting/generate_perf_report_pytorch_inference.py
devalshahamd: I think it is better to keep this since it is less error prone for downstream automation.

TraceLens/Reporting/generate_perf_report_pytorch.py
devalshahamd: Done

kyle-hoffmeyer (approved):

PR #616, opened 2026-04-30
Additional consistency around communication processing

gabeweisz:
More inconsistencies were found in how communication primitives were processed. This PR addresses these including:
- searches for NCCL or RCCL use search instead of match everywhere
- use "nccl" or "rccl" during orphan detection

Still TBD: should try_perf use GPU kernel name via UID lookup or event["name"]?
Should GPUEventAnalyzer/treeperf require correlation id or external id, or should we remove this from NCCL Analyser? Or leave this inconsistent?
Should GPUEventAnalyzer/treeperf allow or even include the custom patterns used by NCCLAnalyser?

### DIFF
--- a/TraceLens/NcclAnalyser/nccl_analyser.py
+++ b/TraceLens/NcclAnalyser/nccl_analyser.py
@@ -8,13 +8,14 @@ import ast
 import gzip
 import os
 import json
-import pandas as pd
-import warnings
-import gzip
 import logging
+import warnings
+
+import pandas as pd
 from concurrent.futures import ProcessPoolExecutor, as_completed
 
 from ..util import DataLoader
+from ..util import DEFAULT_CUSTOM_COLLECTIVE_PATTERNS
 from ..util import TraceEventUtils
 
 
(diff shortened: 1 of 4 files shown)


devalshahamd:
Hi @gabeweisz 

1) Can't think of any advantage/disadvantage of one over other. gpu_event["name"] might be simpler. 
2) I think we can leave that check in GPUEventAnalyzer and TreePerf since it works on linked tree. I think that check was added in NCCL Analyzer since it works on raw trace and not tree. 
3) I think we should use the custom patterns  used by NCCLAnalyser for GPUEventAnalyzer/treeperf as well. So that GPU timeline and perf sheets have consistent categorization.

ajassani:
Replies to the three open questions:

## Q1: Should `try_perf` use GPU kernel name via UID lookup or `event["name"]`?

The perf model lookup should keep using the **CPU op name** (i.e. `event["name"]` on the host op) — that's the right key for `op_to_perf_model_class_map` since perf models are written against CPU op names (ATen ops like `aten::mm` being one example).

The GPU kernel name was only used to detect whether the op is a collective and filter it out (the two spots this PR touches in `generate_perf_report_pytorch_inference.py` and `tree_perf.py` L1826 are both iterating over kernel events specifically for the collective-filter purpose). Let's keep that logic unchanged.

## Q2: Should `GPUEventAnalyzer`/treeperf require correlation id or external id, or should we remove this from `NcclAnalyser`?

The `correlation` / `External id` check in `NcclAnalyser._collective_filter` is a way to confirm that the GPU kernel is linked to a host op in profiling scope — sometimes the profiler records GPU events from preceding or succeeding steps, and we don't want to analyze those.

For `GPUEventAnalyser`/treeperf this isn't needed in the same form because:
- treeperf walks down from CPU-op roots, so the host op is in scope by construction.
- `GPUEventAnalyser` is fed already-filtered kernels via `event.get("tree")` upstream (see `tree_perf.py` L819-820 and L2519-2520) — functionally equivalent to NcclAnalyser's `correlation`/`External id` gate, just a different mechanism (the kernel got linked into the tree).

So the right move is: leave `NcclAnalyser` as-is since it's standalone (no tree to lean on), and don't add the correlation/external-id check to `GPUEventAnalyser`/treeperf.

## Q3: Should `GPUEventAnalyzer`/treeperf allow or even include the custom patterns used by `NcclAnalyser`?

Yes. Goal: one place to register a new pattern (e.g. `quick_reduce`), and `NcclAnalyser`, `GPUEventAnalyser`, and the treeperf collective filter all become aware of it. Today only `NcclAnalyser` honors `custom_collective_patterns`, so workloads with `cross_device_reduce` / `quick_reduce` get misclassified as `computation` in the GPUEventAnalyser breakdown.

devalshahamd (approved):
Looks good to me!

Issue #619, opened 2026-05-04
Add Triton/torch.compile perf models (parity with ATen perf models)


ajassani:
### Background

TraceLens currently has no performance models for `torch.compile`-generated Triton kernels. They show up in the trace as opaque `triton_*` kernels with measured time but no FLOPs, bytes, AI, or achieved-perf — unlike `aten::mm` and other ops that already have perf models and populate `unified_perf_summary`.

Concrete example — `unified_perf_summary` from a small transformer block (RMSNorm + Attention + SwiGLU MLP, B=8, T=4096, dim=2048, bf16) generated through TraceLens today:

| kernel                                               | cat      | kernel_time_us | % gpu | GFLOPS | TFLOPS/s | TB/s | perf model |
|:-----------------------------------------------------|:---------|---------------:|------:|:-------|:---------|:-----|:-----------|
| aten::_scaled_dot_product_flash_attention            | SDPA_fwd |          26637 |  29.0 | 549.8  | 206.4    | 0.20 | yes        |
| aten::mm                                             | GEMM     |          25946 |  28.3 | 755.9  | 583.4    | 0.41 | yes        |
| aten::mm                                             | GEMM     |          15614 |  17.0 | 824.6  | 528.3    | 0.36 | yes        |
| aten::mm                                             | GEMM     |          12216 |  13.3 | 755.9  | 619.2    | 0.43 | yes        |
| aten::mm                                             | GEMM     |           4770 |   5.2 | 274.9  | 576.3    | 0.58 | yes        |
| triton_poi_fused_mul_silu_2                          | triton   |           3241 |   3.5 | —      | —        | —    | no         |
| triton_poi_fused_add_3                               | triton   |           1364 |   1.5 | —      | —        | —    | no         |
| triton_red_fused__to_copy_add_mean_mul_pow_rsqrt_1   | triton   |           1114 |   1.2 | —      | —        | —    | no         |
| triton_red_fused__to_copy_add_mean_mul_pow_rsqrt_0   | triton   |            908 |   1.0 | —      | —        | —    | no         |

ATen rows have full perf metrics; Triton rows are blank. Even on this minimal block ~7% of GPU time lands in kernels we can't characterize; on real workloads with more aggressive fusion the share grows.

> Read the table for *structure* — which rows have perf-model data, which don't. Specific numbers are from a small minimal repro and aren't meant as a benchmark.

#### Repro

[...]

PR #621, opened 2026-05-05
Perfmodel for torch.compile (triton) 

janmatai:
# Background

Issue #619 — Enable perf metrics (GFLOPS, TFLOPS/s, TB/s) for Triton kernels generated by `torch.compile`.

When `torch.compile` fuses ATen ops into Triton kernels, TraceLens previously showed blank metrics for these kernels. This PR adds `TritonCompiledPerfModel` to compute GFLOPS, Data Moved, and throughput metrics.

# Reproducer — `examples/repro_triton_trace.py`

Builds a `TransformerBlock` (RMSNorm + Attention + SwiGLU MLP, `dim=2048`, bf16), compiles with `torch.compile`, and captures a Chrome trace with `torch.profiler`. The trace contains 4 Triton kernels used to validate the perf model.

# Two-Tier Approach

- **V2 (Trace-Intrinsic, primary)** — Extracts metadata directly from Chrome trace event args (`Concrete Inputs`, `Input Dims`, `Input type`). No disk I/O, no cache dependency. Self-contained and portable. Requires PyTorch 2.4+.

- **V1 (Cache-Based, fallback)** — Parses torchinductor cache `.py` files on disk. Used only when V2 fails (older traces lacking required fields). Cache dir can be specified explicitly via `--inductor_cache_dir`.

- **If both fail** — The kernel appears in the report with timing but no perf metrics.

## Approaches Comparison

image

## Future work: where `TORCH_COMPILE_DEBUG=1` adds value

| Feature | File |
|---|---|
| Verify FLOPs formula with exact shapes | `fx_graph_readable.py` |
| Verify byte counting against exact load/store ops | `ir_post_fusion.txt` |
| Graph break detection | `torchdynamo/debug.log` |
| Recompile tracking | `torchdynamo/debug.log` |
| Reproducing kernel issues in isolation | `fx_graph_runnable.py` |

# How to Run

```bash
python -m TraceLens.Reporting.generate_perf_report_pytorch \
    --profile_json_path torch_trace_output_claude/triton_repro/trace.json.gz \
    --output_xlsx_path torch_trace_output_claude/triton_repro/perf_report.xlsx \
    --output_csvs_dir torch_trace_output_claude/triton_repro/csvs_output \
    --enable_pseudo_ops --group_by_num_kernels
```

With explicit inductor cache (for V1 fallback on a different machine):
```bash
python -m TraceLens.Reporting.generate_perf_report_pytorch \
    --profile_json_path trace.json.gz \
    --output_xlsx_path perf_report.xlsx \
    --inductor_cache_dir /path/to/torchinductor/cache
```

# Results

Model: `TransformerBlock(dim=2048, n_heads=16, mlp_ratio=4)`, batch=8, seq_len=4096, dtype=bf16. Traced with PyTorch 2.11+rocm7.2 on MI300X.

| Kernel | GFLOPS | Data Moved (MB) | FLOPS/Byte | TB/s | TFLOPS/s |
|--------|--------|-----------------|------------|------|----------|
| `triton_red_fused_add_mean_mul_pow_rsqrt_0` (RMSNorm) | 0.470 | 256 | 1.75 | 3.48 | 6.10 |
| `triton_red_fused__unsafe_view_add_mean_mul_pow_rsqrt_1` (RMSNorm+residual) | 0.470 | 384 | 1.17 | 4.44 | 5.18 |
| `triton_poi_fused__unsafe_view_mul_silu_2` (SwiGLU) | 1.342 | 1,536 | 0.83 | 3.05 | 2.55 |
| `triton_poi_fused__unsafe_view_add_3` (residual add) | 0.067 | 512 | 0.12 | 3.81 | 0.48 |

# Key Design Decisions (addressing review feedback)

### 1. Broadened mapping API (no Triton-specific logic in `tree_perf.py`)

Added `resolve_perf_model_class()` and `register_perf_model_matcher()` to `torch_op_mapping.py`. `TritonCompiledPerfModel` registers its own match rule. `tree_perf.py` uses a single uniform dispatch path — no prefix checks, no bare `except Exception`. `_has_perf_model()` uses a generic `can_model(event)` protocol so kernels that match the pattern but lack trace-arg metadata are correctly reported as unmodeled. Reusable for any future perf model with generated names.

### 2. Explicit `inductor_cache_dir` kwarg

Added `--inductor_cache_dir` CLI argument following the `arch` / `python_path` / `enable_origami` convention. Threaded through `TreePerfAnalyzer` → `_perf_model_init_kwargs` → `TritonCompiledPerfModel` → `_cache_dirs()`. When provided, takes priority over env var and auto-detected paths. Conditionally passed via signature introspection (same pattern as `enable_origami`).

### 3. No torch import at analysis time

Vendored the op registry as `TraceLens/PerfModel/_known_aten_ops.py` — an auto-generated Python module with 960 op names from torch 2.11.0+rocm7.2. Generated by `scripts/generate_known_aten_ops.py`, refreshed per torch version bump. Uses `.py` (not `.json`) so it's always included in the installed package. `_parse_kernel_name` imports the set at module level — same coverage, deterministic across analysis hosts.

### 4. Regression tests

- **`tests/test_triton_compiled_perf_model.py`** — 6 unit tests covering `_parse_kernel_name`, `_meta_from_trace_args`, `flops()`, and `bytes()` with a RMSNorm kernel. Clear expected values; any change that alters output breaks loudly.
- **`tests/traces/torch_compile_triton/`** — real trace + golden reference CSVs, auto-discovered by `test_perf_report_regression.py`. Validates the full pipeline end-to-end.

# Code Changes Summary

### `tree_perf.py` — uniform dispatch, no Triton-specific logic

- `compute_perf_metrics()`: replaced `op_to_perf_model_class_map.get()` + Triton prefix-check fallback with single call to `resolve_perf_model_class()`
- `_has_perf_model()`: uses `resolve_perf_model_class()` + generic `can_model(event)` check
- `build_df_unified_perf_table()`: removed 20-line opportunistic Triton block + bare `except Exception`; restored simple `row["perf_params"] = None`
- `_perf_model_init_kwargs()`: conditionally passes `inductor_cache_dir` (same signature introspection pattern as `enable_origami`)

### `torch_op_mapping.py` — pattern-based matcher API

```python
_perf_model_matchers: list = []

def register_perf_model_matcher(matcher): ...
def resolve_perf_model_class(name): ...   # dict first, then matchers

def _match_triton_compiled(name):         # lazy import, no torch pulled
    if name.startswith(("triton_poi_", "triton_red_", "triton_per_")):
        from ...triton_compiled_perf_model import TritonCompiledPerfModel
        return TritonCompiledPerfModel
    return None

register_perf_model_matcher(_match_triton_compiled)
```

### `triton_compiled_perf_model.py` — perf model class

- `_parse_kernel_name()`: DP segmentation against vendored op registry (`_known_aten_ops.py`). Fallback to `_FLOPS_PER_ELEM` keys when registry unavailable.
- `_meta_from_trace_args()`: V2 extraction from `Concrete Inputs` / `Input Dims` / `Input type`. Skips non-integer scalars (floats, booleans).
- `_cache_dirs()` / `_lookup()`: V1 cache parsing. Priority: `inductor_cache_dir` kwarg → env var → defaults.
- `can_model()`: static method — returns `True` only if V2 trace args are available and parseable.
- `flops()` / `bytes()`: raises `NotImplementedError` if `_meta` is None (gated by `can_model`).

# Files Changed

| File | Change |
|------|--------|
| `TraceLens/PerfModel/triton_compiled_perf_model.py` | New perf model class (V2 + V1), kernel name parser, `can_model()` |
| `TraceLens/PerfModel/torch_op_mapping.py` | Pattern-based matcher API + Triton registration |
| `TraceLens/PerfModel/_known_aten_ops.py` | Vendored op registry (auto-generated) |
| `TraceLens/PerfModel/__init__.py` | Export `resolve_perf_model_class` |
| `TraceLens/TreePerf/tree_perf.py` | Uniform dispatch via `resolve_perf_model_class`, `inductor_cache_dir` plumbing |
| `TraceLens/Reporting/generate_perf_report_pytorch.py` | `--inductor_cache_dir` CLI arg |
| `scripts/generate_known_aten_ops.py` | Maintenance script for op registry |
| `tests/test_triton_compiled_perf_model.py` | Unit tests |
| `tests/traces/torch_compile_triton/` | Trace + golden reference CSVs |
| `docs/triton_perf_model_walkthrough.md` | Walkthrough documentation |
| `examples/repro_triton_trace.py` | Reproducer script |
| `examples/read_triton_trace_args.py` | Trace inspection utility |

# Known Limitations / Future Improvements

| Item | Description | Priority |
|------|-------------|----------|
| `can_model` only checks V2 | Kernels that could be modeled via V1 (inductor cache) but lack trace args will report `has_perf_model=False`. V1 check would require disk I/O in a simple availability check. | Low — V2 is primary for modern traces |
| `getpass.getuser()` in `_cache_dirs` | Can raise `KeyError` in containers without `$USER`/`/etc/passwd`. Bypassed when `--inductor_cache_dir` is provided. | Low — pre-existing, not introduced by this PR |
| Scalar position assumption | `scalars[-2], scalars[-1]` assumes xnumel/rnumel are always the last integer scalars in `Concrete Inputs`. Needs verification against a wider set of traces. | Medium — investigate in follow-up |
| V1 wrapper regex fragility | `content.find("''', device_str=")` and `"async_compile.triton"` sentinel may break if PyTorch changes Inductor's generated code format. | Low — V1 is fallback only |
| More test cases | Current tests cover V2 + RMSNorm. Follow-ups: V1 wrapper-parsing path, multi-word ops, no-registry fallback. | Medium |


ajassani:
Thanks Janarbek, we discussed two follow-ups that would be good to check before this lands:

1. For `triton_red_fused_add_mean_mul_pow_rsqrt_0`, the perf model reports `256.1 MB`, while the counter-based measured memory traffic is higher. I think this is acceptable and expected: the model is estimating the ideal/algorithmic bytes assuming full memory reuse, while the generated Triton implementation reloads input `X` twice. So the gap here reflects an implementation constraint/inefficiency, not necessarily a perf model bug. It would be good to do the same analysis for `triton_poi_fused__unsafe_view_mul_silu_2`: compare the modeled bytes against measured memory traffic. My hunch is that misaligned strides causing non-coalesced memory access, but good to confirm and understand discrepency.

2. V2 appears to infer fused ops from the Triton kernel name. From my experience, kernel names may not always fully represent the actual fused ops, so can we verify that this naming-based inference holds for Inductor-generated kernels across the cases we care about?

janmatai:
> 2\. V2 appears to infer fused ops from the Triton kernel name. From my experience, kernel names may not always fully represent the actual fused ops, so can we verify that this naming-based inference holds for Inductor-generated kernels across the cases we care about?
Thanks Adeem for the feedback! 

1. Yes, agreed for `triton_red_fused_add_mean_mul_pow_rsqrt_0 ` the `256 MB → ~384 MB` gap is the two-pass reduction reloading X, and we agree this is an expected implementation constraint, not a model bug. For triton_poi_fused__unsafe_view_mul_silu_2, I found the discrepancy was actually a model bug: in_out_ptr0 is both read and written, but we were only counting it once. After fixing this, Data Moved goes from 1024 MB → 1536 MB, and TB/s from 1.78 → 2.67. Same fix applies to triton_poi_fused__unsafe_view_add_3 (384 → 512 MB). Checked all kernels for strides, and it seems the innermost stride is 1 and access is coalesced. 

3. Good point — this is now fixed. The original parser used greedy longest-match against a hardcoded 26-op dictionary, which missed multi-word ops like _unsafe_view and would completely fail on names like `triton_poi_fused_all_to_all_single_clone_permute_0`. The core issue is that underscores serve as both op separators and part of op names, making the segmentation ambiguous. We replaced it with a [DP-based parser](https://github.com/AMD-AGI/TraceLens/blob/janmatai/perf_model4torch_compile_triton/TraceLens/PerfModel/triton_compiled_perf_model.py#L169-L211) that builds its dictionary from torch.ops.aten at import time (~830 ops), plus torch.distributed for collective ops. Inductor sorts ops alphabetically in the name, which helps disambiguate. For our kernels, this now correctly detects _unsafe_view (previously missed).

ajassani:
**The blocking items below are all about *scaffolding so the model can improve over time*, not about hitting 100% accuracy in this PR.** A separate "Accuracy improvements" section captures the substantive accuracy findings the repro surfaced. Those are real but non-blocking, and I'm happy to file each as a follow-up.

## Blocking for merge

### 1. Add a regression test in the style of existing TraceLens perf report regression tests

No `test_*.py` in the PR. The two example scripts aren't tests. Please add a regression test in the style of [`tests/test_perf_report_regression.py`](https://github.com/AMD-AGI/TraceLens/blob/main/tests/test_perf_report_regression.py).

Start with **one** test for a small kernel like RMSNorm (small, clear expected FLOPs/bytes, exercises the V2 path), just to stand up the plumbing. More cases (V1 wrapper-parsing path, multi-word ops, the no-torch L128 fallback) can land in follow-ups. Point of getting one in *this* PR: the next change to `_parse_kernel_name` or `_meta_from_trace_args` breaks loudly instead of silently returning `[]`.

### 2. Perf-model-specific logic shouldn't live in `tree_perf.py`

The PR adds Triton-specific prefix checks (`triton_poi_` / `triton_red_` / `triton_per_`) directly into `tree_perf.py` in two places (L368 in `compute_perf_metrics`, L2061 in the `row["perf_params"] = None` region), plus a bare `except Exception` that "silently leaves metrics blank" (own comment), and that swallow is exactly what hides the accuracy issues below. Knowing which kernels a perf model claims is the perf model's responsibility, not `tree_perf.py`'s.

If the existing mapping API (`op_to_perf_model_class_map`, exact-name keyed) is too narrow for generated kernel names, broaden the mapping API itself. `tree_perf.py` should still see one uniform dispatch path. Then `TritonCompiledPerfModel` registers its own match rule, the special cases disappear from `tree_perf.py`, the bare except goes away (`NotImplementedError` flows through the existing handler), and the same broader API is reusable for any future perf model with generated names.

### 3. Inputs from the capture host should be explicit arguments, not env-var magic

The PR resolves the inductor cache dir via `$TORCHINDUCTOR_CACHE_DIR` → `~/.cache/torchinductor` → `/tmp/torchinductor_{getuser()}`, all of which silently assume same-machine same-user. Captures happen on cluster nodes, analysis may happen on laptops/CI/teammates' boxes. None of those defaults exist there. Add `inductor_cache_dir` as an explicit kwarg matching the `arch` / `python_path` / `enable_origami` convention; env var stays as fallback.

### 4. TraceLens shouldn't import torch at analysis time

The PR does `import torch as _torch` at module load in `triton_compiled_perf_model.py` to build `_ALL_KNOWN_OPS` from `torch.ops.*` + `torch.distributed.*`. From the earlier thread, I see this is deliberate: the new DP parser wants a broad registry (~830 ops from `torch.ops.aten` plus `torch.distributed`), and walking `torch.ops` is the easy way to get it. The coverage goal is right; the means is the problem. It pulls a torch install onto every analysis host (laptops, CI runners, lightweight containers), contradicting the "analyze traces anywhere" promise. (Minor secondary effect: if the analysis host's torch version differs from the capture machine's, the known-op set differs silently, so `_parse_kernel_name` can decode the same trace differently depending on the host.)

The op set is reference data. Treat it like one. Vendor it: a small maintenance script imports torch on a developer machine and dumps the op-name union to a committed JSON (`_known_aten_ops.json`), refreshed per torch version bump. `_parse_kernel_name` reads the JSON at runtime, no torch import. Same coverage as today, deterministic across analysis hosts.

## Accuracy improvements (non-blocking, happy to file as follow-ups)

### Repro

Findings below come from one stress-test kernel: `triton_poi_fused_add_all_reduce_bitwise_and_bitwise_not_bitwise_or_embedding_ge_lt_masked_fill_mul_sub_unsqueeze_0` (the VocabParallelEmbedding kernel from a MiniMax-M2.5 vLLM run, TP=4, MI300X).

```bash
# 1. Serve the model with torch.profiler enabled (capture both a Chrome trace and the inductor cache)
vllm serve MiniMaxAI/MiniMax-M2.5 \
  --tensor-parallel-size 4 --max-model-len 10240 \
  --gpu-memory-utilization 0.95 --quantization fp8 \
  --no-enable-prefix-caching --trust-remote-code

# Workload: any prefill-heavy vLLM load will hit the kernel.
#   In our run: ISL=8192, OSL=1024, concurrency=16, ~160 prompts.
# Torch profiler must be enabled with record_shapes=True (vLLM does this by default
# with VLLM_TORCH_PROFILER_DIR set). Inductor cache lands at ~/.cache/torchinductor
# (or wherever TORCHINDUCTOR_CACHE_DIR points).

# 2. scp the (trace.json.gz, inductor_cache/) pair to the analysis box.

# 3. Point TraceLens at both and grep for the kernel:
export TORCHINDUCTOR_CACHE_DIR=/path/to/copied/inductor_cache
python -m TraceLens.Reporting.generate_perf_report_pytorch \
  --profile_json_path /path/to/rank1.trace.json.gz \
  --output_xlsx_path perf_report.xlsx
# Search the report for triton_poi_fused_add_all_reduce_... and compare to the
# table below.
```

### Findings against this kernel

| | TraceLens reports | Reality |
|---|---:|---:|
| FLOPs/elem | ~4 | ~16 (`bitwise_*`, `ge`/`le`, `masked_fill`, `where`, `unsqueeze` missing or 0 in `_FLOPS_PER_ELEM`) |
| Bytes moved | ~341 MB | ~92 MB (V2 charges full embedding shard `[50016,3072] bf16 = 293 MB` even though only ~2048 rows are gathered) |
| Fused-op list (V1, from wrapper comment) | includes `vllm.all_reduce` | `all_reduce` is a separate `torch.ops.vllm.all_reduce.default` call after the kernel, not fused. Verified by grepping the `.amdgcn`. Orthogonal to the recent DP-parser fix: the parser segments the name correctly, but the name itself contains a non-fused op. |

Most findings collapse into two themes:

#### a) Compose V1 and V2: use the cache to fix what trace args alone can't

Today V2 either succeeds and V1 is never consulted, or V2 returns `None` and V1 takes over. Composition fixes a cluster of problems at once: V2 contributes exact per-tensor shapes, V1 contributes the structured op list and the pointer-direction info from the wrapper signature. Several individual findings collapse into this single change:

- **V2 over-counts bytes on gather/scatter** (the 341 vs 92 MB above). Charges every input tensor at full shape × bpe; for `aten.embedding` only `output_rows` of `vocab_size` are touched. With the fused-op list reliably available from V1, cap that tensor's contribution when the list contains `embedding`/`gather`/`scatter`/`index_select`/`embedding_bag`.
- **V2 can't tell `in_ptr` / `out_ptr` / `in_out_ptr` apart.** The recent `in_out_ptr0` double-count fix in this PR (the kernel 2 / kernel 3 deltas in the WITH2 table) lives in V1's wrapper-signature parsing, which V2 doesn't have. Trace args carry no direction tag, so V2 still counts every listed tensor as a read and blindly adds one extra write for slot 0 on pointwise: spurious double-count when a pointwise kernel has no `in_out_ptr`, missing extra write on reductions that *do* have one. Composition lets V2 borrow V1's `in_out_*` / `in_*` / `out_*` names.
- **Even with the new DP parser, kernel-name parsing is the wrong source of truth.** The DP parser fixes greedy-match misses (`_unsafe_view`, multi-word ops), but the *input* it parses is still a fragile string: segmentation silently returns `[]` whenever Inductor changes its naming convention (already changed once: 2.x → 2.7 moved `size_hints` list → dict). The trace already carries `event["args"]["kernel_file"]` (PT 2.4+) pointing at the wrapper file. Read that directly. (Even better long-term: ask `torch.compile` for a sidecar manifest.)
- **The `# Original ATen:` wrapper comment is misleading.** It's Inductor's *scheduling group*, not the @triton.jit body; `vllm.all_reduce` appears in the comment but is actually emitted as a separate dispatcher call after the kernel. Composition has to cross-check with the call sequence in `output_code.py`, not trust the comment in isolation. At minimum, add a code comment so future contributors don't trust the wrapper comment as fused-op ground truth.

(Framing note: the PR positions V2 as primary and V1 as fallback for older PyTorch. With composition, **trace+cache becomes the accurate default**; trace-only becomes the lighter-but-less-accurate option. Ties to blocker 3: `inductor_cache_dir` needs to be a first-class input for this to work cross-machine.)

#### b) Build the op registry from Inductor's own source of truth, ahead of time

The PR hand-curates two op tables that drift from reality:

- **`_FLOPS_PER_ELEM`** (~25 entries, hand-picked). This kernel alone has ~12 ops missing or zero (`bitwise_*`, `ge`/`le`, `masked_fill`, `where`, `unsqueeze`), and `.get(op, 0)` silently drops them. Asymmetric, too: `lt`/`gt` are in, `le`/`ge` are out.
- **`_ALL_KNOWN_OPS`** is built at runtime by walking `torch.ops.*` + `torch.distributed.*`. Double-violates the no-torch goal (blocker 4) and silently changes shape with the analysis host's torch version.

Both should be derived from **Inductor's own pointwise lowering registry** (`torch._inductor.lowering.register_pointwise`, `lowering.py` [L875](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/lowering.py#L875) / bulk registrations around [L6705](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/lowering.py#L6705)). Dump it once per torch version on a developer machine, commit as JSON. Two wins: (a) the set of valid ops is exactly what Inductor can fuse, no drift; (b) no runtime torch, same registry on every analysis host. (Ties to blocker 4.)

janmatai:
> **The blocking items below are all about _scaffolding so the model can improve over time_, not about hitting 100% accuracy in this PR.** A separate "Accuracy improvements" section captures the substantive accuracy findings the repro surfaced. Those are real but non-blocking, and I'm happy to file each as a follow-up.
> 
> ## Blocking for merge
> ### 1. Add a regression test in the style of existing TraceLens perf report regression tests
> No `test_*.py` in the PR. The two example scripts aren't tests. Please add a regression test in the style of [`tests/test_perf_report_regression.py`](https://github.com/AMD-AGI/TraceLens/blob/main/tests/test_perf_report_regression.py).
> 
> Start with **one** test for a small kernel like RMSNorm (small, clear expected FLOPs/bytes, exercises the V2 path), just to stand up the plumbing. More cases (V1 wrapper-parsing path, multi-word ops, the no-torch L128 fallback) can land in follow-ups. Point of getting one in _this_ PR: the next change to `_parse_kernel_name` or `_meta_from_trace_args` breaks loudly instead of silently returning `[]`.
> 
> ### 2. Perf-model-specific logic shouldn't live in `tree_perf.py`
> The PR adds Triton-specific prefix checks (`triton_poi_` / `triton_red_` / `triton_per_`) directly into `tree_perf.py` in two places (L368 in `compute_perf_metrics`, L2061 in the `row["perf_params"] = None` region), plus a bare `except Exception` that "silently leaves metrics blank" (own comment), and that swallow is exactly what hides the accuracy issues below. Knowing which kernels a perf model claims is the perf model's responsibility, not `tree_perf.py`'s.
> 
> If the existing mapping API (`op_to_perf_model_class_map`, exact-name keyed) is too narrow for generated kernel names, broaden the mapping API itself. `tree_perf.py` should still see one uniform dispatch path. Then `TritonCompiledPerfModel` registers its own match rule, the special cases disappear from `tree_perf.py`, the bare except goes away (`NotImplementedError` flows through the existing handler), and the same broader API is reusable for any future perf model with generated names.
> 
> ### 3. Inputs from the capture host should be explicit arguments, not env-var magic
> The PR resolves the inductor cache dir via `$TORCHINDUCTOR_CACHE_DIR` → `~/.cache/torchinductor` → `/tmp/torchinductor_{getuser()}`, all of which silently assume same-machine same-user. Captures happen on cluster nodes, analysis may happen on laptops/CI/teammates' boxes. None of those defaults exist there. Add `inductor_cache_dir` as an explicit kwarg matching the `arch` / `python_path` / `enable_origami` convention; env var stays as fallback.
> 
> ### 4. TraceLens shouldn't import torch at analysis time
> The PR does `import torch as _torch` at module load in `triton_compiled_perf_model.py` to build `_ALL_KNOWN_OPS` from `torch.ops.*` + `torch.distributed.*`. From the earlier thread, I see this is deliberate: the new DP parser wants a broad registry (~830 ops from `torch.ops.aten` plus `torch.distributed`), and walking `torch.ops` is the easy way to get it. The coverage goal is right; the means is the problem. It pulls a torch install onto every analysis host (laptops, CI runners, lightweight containers), contradicting the "analyze traces anywhere" promise. (Minor secondary effect: if the analysis host's torch version differs from the capture machine's, the known-op set differs silently, so `_parse_kernel_name` can decode the same trace differently depending on the host.)
> 
> The op set is reference data. Treat it like one. Vendor it: a small maintenance script imports torch on a developer machine and dumps the op-name union to a committed JSON (`_known_aten_ops.json`), refreshed per torch version bump. `_parse_kernel_name` reads the JSON at runtime, no torch import. Same coverage as today, deterministic across analysis hosts.
> 
> ## Accuracy improvements (non-blocking, happy to file as follow-ups)
> ### Repro
> Findings below come from one stress-test kernel: `triton_poi_fused_add_all_reduce_bitwise_and_bitwise_not_bitwise_or_embedding_ge_lt_masked_fill_mul_sub_unsqueeze_0` (the VocabParallelEmbedding kernel from a MiniMax-M2.5 vLLM run, TP=4, MI300X).
> 
> ```shell
> # 1. Serve the model with torch.profiler enabled (capture both a Chrome trace and the inductor cache)
> vllm serve MiniMaxAI/MiniMax-M2.5 \
>   --tensor-parallel-size 4 --max-model-len 10240 \
>   --gpu-memory-utilization 0.95 --quantization fp8 \
>   --no-enable-prefix-caching --trust-remote-code
> 
> # Workload: any prefill-heavy vLLM load will hit the kernel.
> #   In our run: ISL=8192, OSL=1024, concurrency=16, ~160 prompts.
> # Torch profiler must be enabled with record_shapes=True (vLLM does this by default
> # with VLLM_TORCH_PROFILER_DIR set). Inductor cache lands at ~/.cache/torchinductor
> # (or wherever TORCHINDUCTOR_CACHE_DIR points).
> 
> # 2. scp the (trace.json.gz, inductor_cache/) pair to the analysis box.
> 
> # 3. Point TraceLens at both and grep for the kernel:
> export TORCHINDUCTOR_CACHE_DIR=/path/to/copied/inductor_cache
> python -m TraceLens.Reporting.generate_perf_report_pytorch \
>   --profile_json_path /path/to/rank1.trace.json.gz \
>   --output_xlsx_path perf_report.xlsx
> # Search the report for triton_poi_fused_add_all_reduce_... and compare to the
> # table below.
> ```
> 
> ### Findings against this kernel
> TraceLens reports	Reality
> FLOPs/elem	~4	~16 (`bitwise_*`, `ge`/`le`, `masked_fill`, `where`, `unsqueeze` missing or 0 in `_FLOPS_PER_ELEM`)
> Bytes moved	~341 MB	~92 MB (V2 charges full embedding shard `[50016,3072] bf16 = 293 MB` even though only ~2048 rows are gathered)
> Fused-op list (V1, from wrapper comment)	includes `vllm.all_reduce`	`all_reduce` is a separate `torch.ops.vllm.all_reduce.default` call after the kernel, not fused. Verified by grepping the `.amdgcn`. Orthogonal to the recent DP-parser fix: the parser segments the name correctly, but the name itself contains a non-fused op.
> Most findings collapse into two themes:
> 
> #### a) Compose V1 and V2: use the cache to fix what trace args alone can't
> Today V2 either succeeds and V1 is never consulted, or V2 returns `None` and V1 takes over. Composition fixes a cluster of problems at once: V2 contributes exact per-tensor shapes, V1 contributes the structured op list and the pointer-direction info from the wrapper signature. Several individual findings collapse into this single change:
> 
> * **V2 over-counts bytes on gather/scatter** (the 341 vs 92 MB above). Charges every input tensor at full shape × bpe; for `aten.embedding` only `output_rows` of `vocab_size` are touched. With the fused-op list reliably available from V1, cap that tensor's contribution when the list contains `embedding`/`gather`/`scatter`/`index_select`/`embedding_bag`.
> * **V2 can't tell `in_ptr` / `out_ptr` / `in_out_ptr` apart.** The recent `in_out_ptr0` double-count fix in this PR (the kernel 2 / kernel 3 deltas in the WITH2 table) lives in V1's wrapper-signature parsing, which V2 doesn't have. Trace args carry no direction tag, so V2 still counts every listed tensor as a read and blindly adds one extra write for slot 0 on pointwise: spurious double-count when a pointwise kernel has no `in_out_ptr`, missing extra write on reductions that _do_ have one. Composition lets V2 borrow V1's `in_out_*` / `in_*` / `out_*` names.
> * **Even with the new DP parser, kernel-name parsing is the wrong source of truth.** The DP parser fixes greedy-match misses (`_unsafe_view`, multi-word ops), but the _input_ it parses is still a fragile string: segmentation silently returns `[]` whenever Inductor changes its naming convention (already changed once: 2.x → 2.7 moved `size_hints` list → dict). The trace already carries `event["args"]["kernel_file"]` (PT 2.4+) pointing at the wrapper file. Read that directly. (Even better long-term: ask `torch.compile` for a sidecar manifest.)
> * **The `# Original ATen:` wrapper comment is misleading.** It's Inductor's _scheduling group_, not the @triton.jit body; `vllm.all_reduce` appears in the comment but is actually emitted as a separate dispatcher call after the kernel. Composition has to cross-check with the call sequence in `output_code.py`, not trust the comment in isolation. At minimum, add a code comment so future contributors don't trust the wrapper comment as fused-op ground truth.
> 
> (Framing note: the PR positions V2 as primary and V1 as fallback for older PyTorch. With composition, **trace+cache becomes the accurate default**; trace-only becomes the lighter-but-less-accurate option. Ties to blocker 3: `inductor_cache_dir` needs to be a first-class input for this to work cross-machine.)
> 
> #### b) Build the op registry from Inductor's own source of truth, ahead of time
> The PR hand-curates two op tables that drift from reality:
> 
> * **`_FLOPS_PER_ELEM`** (~25 entries, hand-picked). This kernel alone has ~12 ops missing or zero (`bitwise_*`, `ge`/`le`, `masked_fill`, `where`, `unsqueeze`), and `.get(op, 0)` silently drops them. Asymmetric, too: `lt`/`gt` are in, `le`/`ge` are out.
> * **`_ALL_KNOWN_OPS`** is built at runtime by walking `torch.ops.*` + `torch.distributed.*`. Double-violates the no-torch goal (blocker 4) and silently changes shape with the analysis host's torch version.
> 
> Both should be derived from **Inductor's own pointwise lowering registry** (`torch._inductor.lowering.register_pointwise`, `lowering.py` [L875](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/lowering.py#L875) / bulk registrations around [L6705](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/lowering.py#L6705)). Dump it once per torch version on a developer machine, commit as JSON. Two wins: (a) the set of valid ops is exactly what Inductor can fuse, no drift; (b) no runtime torch, same registry on every analysis host. (Ties to blocker 4.)

Hi, @ajassani ,

I believe I addressed the most of the (focusing the merge blocking ones) comments. See below: 

### 1. Regression test

Added a regression test for RMSNorm (reduction kernel, bf16, V2 path):

- **`tests/test_triton_compiled_perf_model.py`** — 6 unit tests covering `_parse_kernel_name`, `_meta_from_trace_args`, `flops()`, and `bytes()`. Clear expected values; any change that alters output breaks loudly.
- **`tests/traces/torch_compile_triton/`** — real trace + golden reference CSVs, auto-discovered by the existing `test_perf_report_regression.py`. Validates the full pipeline end-to-end.

More cases (V1 path, multi-word ops, no-torch fallback) can land in follow-ups.

### 2. Triton logic out of `tree_perf.py`

Broadened the mapping API in `torch_op_mapping.py` to support pattern-based matchers:

- Added `resolve_perf_model_class(name)` — checks the exact-name dict first, then iterates registered matchers
- `TritonCompiledPerfModel` registers its own match rule via `register_perf_model_matcher()`
- `tree_perf.py` uses `resolve_perf_model_class` everywhere — both prefix-check blocks and the bare `except Exception` are removed
- `_has_perf_model()` goes through the same API, with a generic `can_model(event)` check so kernels that match the pattern but lack trace-arg metadata are correctly reported as unmodeled
- Triton class is lazy-imported to avoid pulling torch at analysis time

### 3. Explicit `inductor_cache_dir`

Added `inductor_cache_dir` as an explicit kwarg following the `arch` / `python_path` / `enable_origami` convention:

- `--inductor_cache_dir` CLI argument added to `generate_perf_report_pytorch.py`
- Threaded through `TreePerfAnalyzer` → `_perf_model_init_kwargs` → `TritonCompiledPerfModel` → `_cache_dirs()`
- When provided, takes priority over all defaults; env var and auto-detected paths remain as fallback
- Conditionally passed via signature introspection (same pattern as `enable_origami`)

### 4. No torch import at analysis time

Vendored the op set as committed reference data:

- **`scripts/generate_known_aten_ops.py`** — maintenance script that dumps the op-name union (`torch.ops.*` + `torch.distributed`) to a Python module
- **`TraceLens/PerfModel/_known_aten_ops.py`** — auto-generated set of 960 ops from torch 2.11.0+rocm7.2, refreshed per torch version bump. Uses `.py` (not `.json`) so it's always included in the installed package without `package_data` configuration.
- `triton_compiled_perf_model.py` imports the set at module level instead of importing torch — same coverage, deterministic across analysis hosts

docs/triton_perf_model_walkthrough.md
```diff
@@ -0,0 +1,573 @@
...
+    activities=[
+        torch.profiler.ProfilerActivity.CPU,
+        torch.profiler.ProfilerActivity.CUDA,
+    ],
+    record_shapes=True,   # required for V2 metadata
+) as prof:
+    model(inputs)
+```
+
+Without `record_shapes=True`, the trace will not contain `Concrete Inputs`,
```
ajassani, line 79: we dont support perf model without record_shapes=True anyways so no need to support for torch compile as well

ajassani (approved):

PR #622, opened 2026-05-05
Add op category breakdown to perf report comparison (#331)

### LINKED ISSUE #331: Add op category to perf report comparison

ajassani:
## Summary

Closes #331.

The per-report flow already produces an `ops_summary_by_category` sheet (via `TreePerfAnalyzer.get_df_kernel_launchers_summary_by_category`), but `compare_perf_reports_pytorch` never picked it up — so when comparing two reports there was no per-category breakdown of how kernel time shifted. This PR wires that sheet into the comparator.

- Add `SHEETS_COMPARE_CONFIG["ops_summary_by_category"]` — key `op category`, diff cols `total_direct_kernel_time_ms` and `Count`, sort by `total_direct_kernel_time_ms`. The µs-scale `_sum` column is already dropped upstream so no extra cleanup is needed.
- New block in `generate_compare_perf_reports_pytorch` parallel to the `ops_summary` one: silently skip the sheet when the source report lacks it (older reports), raise if the user asked for it explicitly.
- Add `ops_summary_by_category` to the `--sheets` CLI choices.

`kernel_summary` is intentionally left alone — `Parent op category` continues to be stripped before the merge as before.

## Test plan

- [x] `pytest tests/test_compare_perf_reports.py` — 5 passed (existing reference fixtures don't include the new sheet but `_validate_report_against_reference` only iterates over reference sheets, so they still pass)
- [x] Manual e2e on the two test traces under `tests/traces/compare_test_e2e/` — confirmed the source reports both include `ops_summary_by_category` and that the comparator now writes a non-empty `ops_summary_by_category.csv` with the expected `<tag>::*` baseline/variant columns and `*__<tag>_diff` / `*__<tag>_pct` diff columns sorted by `total_direct_kernel_time_ms`
- [ ] Reviewer: re-run `tests/test_compare_perf_reports.py`
- [ ] Reviewer: optional — regenerate `tests/traces/compare_test_e2e/reference/compare_h100_mi300_qwen_csvs/` to include `ops_summary_by_category.csv` so the e2e reference test starts validating the new sheet (happy to do as a follow-up)

## Notes for reviewer

- No new dependency, no new public function — only config + a small parallel block. Behavior on older reports that lack `ops_summary_by_category` is unchanged (sheet silently absent under `--sheets all`).
- Optional follow-up: a tiny unit test in `tests/test_compare_perf_reports.py` that builds two synthetic CSV report dirs and asserts the diff/pct columns. Skipped here to keep the PR small; happy to add if you'd like.


TraceLens/Reporting/compare_perf_reports_pytorch.py
```diff
@@ -430,6 +442,19 @@ def generate_compare_perf_reports_pytorch(
...
         results[sheet_to_load] = ops
 
+    # Perform ops_summary_by_category if specified (issue #331).
+    # Older reports won't have this sheet, so we silently skip when missing
+    # under "all"; explicit selection still raises so the user sees the issue.
+    if "ops_summary_by_category" in sheets or "all" in sheets:
+        sheet_to_load = "ops_summary_by_category"
+        if sheet_to_load in report_sheet_names:
+            config = SHEETS_COMPARE_CONFIG[sheet_to_load]
+            results[sheet_to_load] = process_summary_sheet(
```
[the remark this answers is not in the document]
ajassani: Good catch on the comment over-promising — pushed 75b16365 to drop those two lines. The actual existence check intentionally matches the existing ops_summary / kernel_summary blocks (only validates reports[0]); the assumption across this whole comparator is that the inputs come from compatible TraceLens versions. Tightening the missing-sheet detection across all reports would be a worthwhile follow-up but applies symmetrically to all three blocks, so leaving it as a separate cleanup.

gabeweisz (approved):

TraceLens/Reporting/compare_perf_reports_pytorch.py
```diff
@@ -67,6 +67,18 @@
         "cols_to_delete": ["total_direct_kernel_time_sum"],
         "sort_col": "total_direct_kernel_time_ms",
     },
+    # Per-category aggregation produced by
```
kyle-hoffmeyer, line 70: Is this comment block needed? The other sheet configs don't have a comment block

TraceLens/Reporting/compare_perf_reports_pytorch.py
```diff
@@ -430,6 +442,17 @@ def generate_compare_perf_reports_pytorch(
         ops = process_summary_sheet(reports, sheet_to_load, tags, config)
         results[sheet_to_load] = ops
 
+    # Perform ops_summary_by_category if specified (issue #331).
```
kyle-hoffmeyer, line 445: The code structure for this new section seems to be a little different from the other sheets (ops_summary, kernel_summary). Is there a reason for this? It seems to be functionally the same though

kyle-hoffmeyer (approved):
Looks good to me! Some small formatting comments but nothing major

PR #634, opened 2026-05-07
Move file to correct location

gabeweisz:
Move file to correct location


gabeweisz:
Not needed - policy allows this file in .github but the checker was wrong

PR #652, opened 2026-05-19
[Feature] Integrating Comparative Analysis into TraceLens Agent

kyle-hoffmeyer:
## Summary

This PR adds a **comparative analysis mode** to the TraceLens Agent. The existing standalone analysis analyzes a single trace in isolation. Comparative mode accepts two traces, a baseline (Trace 1) and a target (Trace 2) trace, and produces a cross-trace report that identifies where and by how much the baseline trace is slower than the target trace.<br>

### How does it work?

The analysis relies on TraceDiff to group kernels by the least common ancestor (LCA) between two traces. Assume that the call stacks of the traces are structured as below: 

```mermaid
graph TD
  subgraph T1["Trace 1 (baseline)"]
    direction TB
    A1[nn.Module: Linear] --> B1[aten::addmm]
    A1 --> C1[aten::transpose]
    B1 --> K1[Cijk_Alik_Bljk]
    C1 --> L1["void at::native::elementwise..."]
  end

  subgraph T2["Trace 2 (target)"]
    direction TB
    A2[nn.Module: Linear] --> B2[aten::linear]
    A2 --> C2[aten::transpose]
    B2 --> K2[sm90_xmma_gemm]
    C2 --> L2["void at::native::elementwise..."]
  end
```

In the example above `Cijk_Alik_Bljk` and `sm90_xmma_gemm` are considered to be the same functional kernel as they align structurally. When performing cross-trace performance analysis, the performance of `Cijk_Alik_Bljk` is then compared to `sm90_xmma_gemm`.

As the new comparative mode is **integrated** into the existing standalone analysis, this new analysis is invoked through the same skill file, but through different prompts:

`"Follow the analysis orchestrator installed with TraceLens and run the full agentic analysis workflow on <path_to_trace.json>"` -> invokes standalone analysis<br>
`"Follow the analysis orchestrator installed with TraceLens and run the full agentic analysis workflow on <path_to_trace1.json> and <path_to_trace2.json>"` -> invokes comparative analysis

### Standalone vs Comparative: Key Differences
| | Standalone | Comparative |
|---|---|---|
| **Inputs** | One trace, one platform | Two traces, two platforms |
| **Efficiency metric** | Distance to roofline peak | Trace 1 performance relative to Trace 2 performance — no roofline involved |
| **Impact** | Projected improvement using a hardware performance model | Runtime difference between the two traces |
| **Fusion candidates** | Identified using key patterns | Identified from TraceDiff. |

### Analysis scope

[...]

PR #654, opened 2026-05-19
Changes to TraceLens files for comparative analysis

kyle-hoffmeyer:
## This PR supplements https://github.com/AMD-AGI/TraceLens/pull/652. All changes in this PR are included in https://github.com/AMD-AGI/TraceLens/pull/652.

**NOTE: CSV reference reports have been updated. Only changes to reference reports are intentional changes as a result of this PR**

`trace_diff.py`:
Added new columns to diff_stats output:
- Added `gpu_op_uid` to diff_stats output. Allows for trace1↔trace2 kernel matching downstream
- Added GPU `busy_time` for each LCA group. Allows for trace1 LCA time and trace2 LCA time comparison downstream

`tree_perf.py`:
- Adds kernel time to `unified_perf_summary` if perf model hits `Exception`.
- Adds GPU op uids to `unified_perf_summary` to be used in comparative analysis. Closes https://github.com/AMD-AGI/TraceLens/pull/624. 

`generate_perf_report_pytorch.py` and `generate_perf_report_pytorch_inference.py`:
- Can accept extension with a function called `postprocess_perf_report_dataframes_extension`. 
  - Previously it could only accept extension files that added sheets to perf report through `get_additional_dataframes_extension`. Now it can accept extension files that modify sheets in perf report through `postprocess_perf_report_dataframes_extension`
- Accepts `extension_args` flag. This is passed into extension file if extension file accepts this flag

`tracediff_comparison_extension.py`:
- Used in comparative analysis to add TraceDiff output (diff_stats.csv) to perf report and modify `unified_perf_summary` to include trace2 times for each row
- More details can be found here: https://github.com/AMD-AGI/TraceLens/pull/652


ajassani:
Since this extension is implementing a TraceLens-owned comparison workflow, can we make the comparison trace a first-class argument instead of passing it through generic `extension_args`?

Something like `--comparison_trace_json_path` could trigger the TraceDiff perf-report enrichment directly. That would make the feature discoverable from `--help`, avoid an untyped string protocol like `"trace2.json debug"`, and remove the need for users to know which extension file to pass for this built-in workflow.

ajassani:
`kernel_details_summary` is meant to stay summary-shaped, but `gpu_op_uids` is an enumeration/provenance list. For grouped rows this can become large and noisy, and it changes the semantics of the summary object. I think having an example UID field like `ex_uid` in each kernel summary entry is fine for traceability, but embedding the full UID list here feels too much.

ajassani:
The fallback to measured GPU busy time when perf metric computation fails makes sense to me. The current implementation duplicates the fallback logic across a few branches; not blocking this PR on it, but we should probably clean that up later with a small helper so the perf-model failure and no-perf-model paths stay consistent.

ajassani:
Thinking more about the UID provenance: if the comparison workflow needs unsummarized GPU event identity, could we do that join before `unified_perf_summary` is created? `generate_perf_report_pytorch.py` already builds `df_unified_perf` first, then calls `summarize_df_unified_perf_table(...)`. We can add the UID in `kernel_details` on the unsummarized `df_unified_perf`, where each concrete GPU event still naturally belongs, and let the TraceDiff/comparison enrichment happen there before summarization. That would avoid embedding full `gpu_op_uids` enumerations inside `kernel_details_summary`, which feels like summary output carrying internal join state.

kyle-hoffmeyer:
> Thinking more about the UID provenance: if the comparison workflow needs unsummarized GPU event identity, could we do that join before `unified_perf_summary` is created? `generate_perf_report_pytorch.py` already builds `df_unified_perf` first, then calls `summarize_df_unified_perf_table(...)`. We can add the UID in `kernel_details` on the unsummarized `df_unified_perf`, where each concrete GPU event still naturally belongs, and let the TraceDiff/comparison enrichment happen there before summarization. That would avoid embedding full `gpu_op_uids` enumerations inside `kernel_details_summary`, which feels like summary output carrying internal join state.

Yes this makes sense, thanks for the clarification. Since we won't be relying on the traditional `extension_file` flag post-summarization, this should work

TraceLens/TreePerf/tree_perf.py
```diff
@@ -1317,16 +1318,15 @@ def _summarize_kernel_stats(series_of_kernel_lists, agg_metrics=["mean"]):
...
         summary_list = copy.deepcopy(template)
 
         # Now, compute statistics and populate the summary list
         for i, kernel_summary in enumerate(summary_list):
             durations_for_this_index = all_durations[i]
             dur_arr = np.array(durations_for_this_index)
 
-            # --- CHANGE: Use consistent naming and clear up original 'dur' key ---
             del kernel_summary["dur"]
+            kernel_summary.pop("gpu_op_uid", None)
```
ajassani, line 1329: why we need this pop?

ajassani (approved):

TraceLens/TreePerf/tree_perf.py
kyle-hoffmeyer: It was added to the kernel_details dictionary to be used by the extension. By the time _summarize_kernel_stats is called, that field is no longer needed and can be popped. Similar to the "dur" field.

ajassani (approved):

ajassani (approved):

PR #660, opened 2026-05-26
Add gpu_op_uid and busy_time columns to TraceDiff output

kyle-hoffmeyer:
This PR adds two columns to TraceDiff output to support comparative analysis.

`trace_diff.py`:
Added new columns to diff_stats output:
- Added `gpu_op_uid` to diff_stats output. Allows for trace1↔trace2 kernel matching downstream
- Added GPU `busy_time` for each LCA group. Allows for trace1 LCA time and trace2 LCA time comparison downstream


spandoesai (approved):
LGTM! Only changes are adding a few fields to the output, sorting the output based on the new fields and a sanity check for filtering events that do not have a mapped uid

PR #677, opened 2026-05-29
Classify graph-replay synthetic ops by GPU kernel name

rkarhila-amd:
## Summary

- Extend `categorize_torch_op()` so TreePerf placeholder ops (`{parent}->{kernel} (Synthetic Op)`) are categorized from the underlying GPU kernel via `classify_kernels`, not the graph-launch parent name.
- Map CK/AITER kernels to TraceLens buckets (`GEMM`, `MoE_unfused`, `GroupQuant`, `RMSNorm`, `elementwise`, `MoE_aux`, etc.) instead of `other`.
- Add unit tests in `tests/test_synthetic_op_kernel_categorization.py`.

## Sample output 

Analysis of SGLang Minimax 2.5 traces. Before this fix, everything launched in graphs was categorised as "other" and no rooflines are provided. After the fix, kernels in graphs are categorised correctly and rooflines and potential improvements are calculated.

### Top Operations without fix on main branch 2026/06/08:
| Rank | Category | Time (ms) | % of Compute Time | Ops |
|------|----------|-----------|-------------------|-----|
| 1 | other | 5995.619 | 98.50% | 498 |

### Top Operations with this fix
| Rank | Category | Time (ms) | % of Compute Time | Ops | Potential improvement (time, E2E %) |
|------|----------|-----------|-------------------|-----|-------------------------------------|
| 1 | GEMM | 1283.434 | 21.1% | 14 | ~652.8–870.9 ms (8.5–11.3%) |
| 2 | elementwise | 1290.999 | 21.2% | 505 | ~621.2–827.8 ms (8.1–10.7%) |
| 3 | MoE_unfused | 1275.301 | 21.0% | 6 | ~507.9–676.7 ms (6.6–8.8%) |
| 4 | GroupQuant | 571.695 | 9.4% | 12 | ~426.2–568.0 ms (5.5–7.4%) |
| 5 | RMSNorm | 568.149 | 9.3% | 10 | ~423.1–564.2 ms (5.5–7.3%) |
| 6 | MoE_aux | 457.841 | 7.5% | 6 | ~341.4–455.5 ms (4.4–5.9%) |
| 7 | SDPA_fwd | 451.240 | 7.4% | 2 | ~238.2–317.5 ms (3.1–4.1%) |

### Smoketest on vllm

Top operations from analysis of a small Qwen3-0.6B vllm trace with graphs. Kernels in graphs are found and rooflines are estimated.

| Rank | Category | Time (ms) | % of Compute Time | Ops | Potential improvement (time, E2E %) |
|------|----------|-----------|-------------------|-----|-------------------------------------|
| 1 | InferenceAttention | 7.093 | 24.9% | 16 | ~5.1–6.9 ms (1.4–1.9%) |
| 2 | GEMM | 9.533 | 33.5% | 39 | ~3.2–4.2 ms (0.9–1.2%) |
| 3 | triton | 6.937 | 24.4% | 79 | no roofline |
| 4 | other | 1.710 | 6.0% | 27 | no roofline |

## Context

Graph-replay traces often attribute GPU work under `hipGraphLaunch` / `cudaGraphLaunch` without a normal `aiter::` / `aten::` CPU op. This is the open-source counterpart to TraceLens-internal #458.

## Regression test updates

Refreshed checked-in CSV golden files (commit `23e80f9a`) after intentional reporting changes from synthetic-op categorization and new graph-replay perf models. All 43 targeted regression cases now pass locally.

| Test module | Cases updated | Why references changed |
|-------------|---------------|------------------------|
| `test_perf_report_regression.py` | 12 traces (h100 × 6, mi300 × 5, `perf_model/normalization`) | `Parent op category`: `other` → `SDPA_fwd` / `elementwise`; `has_perf_model`: `False` → `True`; normalization trace row order + RMSNorm `perf_params` (ScalarList dtype fallback) |
| `test_compare_perf_reports.py` | `compare_test_e2e` h100 + mi300 Qwen references | Category rollups and intersect sheets reflect SDPA/elementwise split out of `other` |
| `test_inference_perf_report.py` | `graph_full`, `graph_piecewise` | Graph-replay MoE aux + elementwise synthetic ops leave `other`; `has_perf_model` flags set |
| `test_detect_recompute.py` | `mi300/resnet_act_checkpoint` | `unified_perf_summary` row order shifts when grouping keys change |

**Trace directories touched (62 CSV files):** `h100/*` (6 workloads), `mi300/*` (5 workloads + recompute), `compare_test_e2e/reference/*`, `inference/graph_full`, `inference/graph_piecewise`, `perf_model/normalization`.

**Unchanged (still pass):** `mi210/gpu_only_trace`, `torch_compile_triton/trace`.

## Test plan

- [x] `pytest tests/test_synthetic_op_kernel_categorization.py -v`
- [x] `pytest tests/test_perf_report_regression.py tests/test_compare_perf_reports.py tests/test_inference_perf_report.py tests/test_detect_recompute.py`
- [ ] Re-run inference perf report + orchestrator on a graph-replay + capture trace and confirm `other` bucket shrinks and `gemm_ops.csv` / `moe_unfused_ops.csv` populate


rkarhila-amd:
The regression test fails because in earlier version any operations inside cuda graph were categorised as "other" whereas this PR gives them explicit categorisation.


| This PR  | Test specification |
|----------|-----------|
| SPDA_fwd | other |
| elementwise | other |

This PR cannot pass the test as it makes a more fine grained analysis. We need this feature implemented in this way or another. Codeowners please comment @ajassani @gabeweisz @AMD-AGI/perf-opt

gabeweisz:
> The regression test fails because in earlier version any operations inside cuda graph were categorised as "other" whereas this PR gives them explicit categorisation.
> 
> This PR	Test specification
> SPDA_fwd	other
> elementwise	other
> This PR cannot pass the test as it makes a more fine grained analysis. We need this feature implemented in this way or another. Codeowners please comment @ajassani @gabeweisz @AMD-AGI/perf-opt

The thing to do in this case is to update the reference outputs. We will be able to see how they changed in the deltas in the CSVs and make sure that they make sense.

I think we have a script that you can use to update the reference results. If not, you can manually find the right files or ask Cursor or whichever tool you like to do it for you

TraceLens/PerfModel/torch_op_mapping.py
```diff
@@ -229,6 +231,53 @@
...
+        warnings.warn(
+            "WARNING: classify_kernel could not confidently classify a GPU kernel; "
+            "categorize_torch_op will fall back to 'other'. "
+            f"kernel_name={kernel_name[:200]!r}",
+            UserWarning,
+            stacklevel=2,
+        )
+        return None
+    ktl, knl = kt.lower(), kernel_name.lower()
+    if kt == "KV Cache Store":
```
gabeweisz: I'd prefer to see a loop over a dict rather than this long list of if statements - I think that is a bit more maintainable

rkarhila-amd: Agreed, mappings should be in same format as for the non-graph kernels.

While working on this, I understood that the suggested change only affects reporting, and perf models are not applied to the kernels in graphs.

I'll see how massive overhaul it takes to get that bit working.

rkarhila-amd:
Turned out to be quite an overhaul. Currently developed with only a single trace. I'll test with a few more cases tomorrow.

rkarhila-amd:
The PR affects almost 80 files but majority of them are test cases.

PR covers now the kernel types inside graphs that were listed in PR description chart. Still tested only on sglang,
I'll try to find useful vLLM task to see how badly it breaks on that side.

rkarhila-amd:
This looks like its working satisfactorily now. I'm running a final end-to-end test on sglang side. Still missing vllm test case, I'll get it tomorrow.

FYI @devalshahamd I had an improperly activated patch in last traces, now I'm testing with properly instrumented Kimi-K2.6-MXFP4 traces accessible to AMDeers in [sharepoint](https://amdcloud.sharepoint.com/:f:/s/SiloEngineering/IgCDvXQPb4l4S4QPIEiUqh3DAV0WEMZR4Kh1UusVWFQog3Q?e=BBPMtY).

rkarhila-amd:
Finally verified that these changes didn't break vllm trace handling.

Please review now.

If this is too much in one go, I can try to break it up to parts, for example I put considerable work on (directing cursor to) optimize the perf tree building in trace merging, and reduced the compute time for some of the operations from hours to seconds. (No matter how I try to avoid it, I somehow always find my way back to tweaking transducers and graph operations). That could be its own PR. The rest is difficult to split.

tkarttunen:
@rkarhila-amd 

### Status audit from a downstream consumer (inference-testing)

Context: ITT has pinned `feature/kernel-synthetic-op-categorization` since Jun 29
(ITT #488) and is now moving to `release/hyperloom_integration_v1.0`. While making
that call we audited this branch against `main` and the v1.0 release, and the
findings seem relevant to what's worth keeping here.

#### Still unique to this PR — no equivalent on `main` or v1.0

- **`TraceLens/Trace2Tree/trace_sglang_capture_link.py`** (564 lines, with a
  333-line test suite). Links SGLang graph-replay kernels to `bs_<N>_rank0.json.gz`
  capture traces by positional alignment (replay kernel *i* ↔ capture launch
  *i + offset*, with `detect_launch_offset` probing up to `MAX_OFFSET_SEARCH = 6`)
  and copies `Input Dims` / `Input type` / `Input Strides` / `Concrete Inputs`
  across. Also carries heuristic shape reconstruction for CK MoE stage-1/stage-2
  layouts and group-size inference from neighbouring quant ops, plus a
  `DECODE_STEP_RE` that already tolerates roofline aggregates folded into the
  step span (`step[DECODE bs=64 g_sqsq=... ]`).
- **Synthetic-op categorization end to end.** Grepping `synthetic_op|enrich_synthetic|graph_replay`
  across `TraceLens/*.py` returns 0 files on `main` and 4 on this branch.
- **Graph-replay roofline for MoE GEMM, MoE_aux, SDPA and RMSNorm** (0ec58d2d, 5b48a402).
- **`event_time_index.py`** (79 lines), and the `bs_<N>_rank0` / `launch_offset`
  concepts, which exist nowhere else in the repo.

#### Overtaken by `main` — likely droppable on the next rebase

- Annotation parsing: `main` landed `TraceUtils/annotation_utils.py` in #841
  (Jul 31) and this branch inherited it through the Aug 6 merge, so nothing
  distinctive remains on that front.
- `extensions/pseudo_ops_registry.py` originates from `main` #869 (cyclic-import
  fix); the branch adds ~125 lines of synthetic-op categories on top of it.
- The capture-merge path has moved considerably since this branch started:
  #707 (graph capture + tracediff), #795 (kernel-name matching for multistream),
  #798 (robustness to graph/capture misordering), #861 (memset/buffer-fill matching).

#### How the two paths relate

`trace_sglang_capture_link` is **not** an alternative to `merge_capture_trace_into_graph`
— it imports `find_closest_batch_size` from it and runs afterwards. The synthetic-op
builder in `tree_perf.py` only visits GPU events not already in `collected_gpu_uids`,
splitting them into true orphans (no `cpu_op` ancestor, i.e. graph replay) and
leftovers that have one. So this PR is **residual coverage layered on the merge
path**, not a replacement for it. Stating that explicitly in the PR description
would help reviewers, since the overlap in problem domain reads as competition.

That framing makes the value question empirical: *what fraction of GPU kernels does
the release leave without an owning op or roofline class?* As a datapoint, Hyperloom
v1.0.0a3 pins the release (`c3405111`) and performs graph-replay categorization and
rooflining successfully via the merge path plus the pre-existing pseudo-op extensions,
so the common case is already covered. What this PR adds is whatever remains, and a
before/after count on one representative MoE trace would size it decisively.

#### Divergence cost

The branch sat untouched from Jul 7 to Aug 6, then took a `main` merge (476bb7c0)
that required a repair commit (8dcaed5f). It is currently 16 ahead / 12 behind `main`,
with a library-code delta of 46 files, +3016/−442. That gap keeps widening while `main`
reworks the same modules, and each repair lands unreviewed. Splitting out a smaller PR
limited to capture-link plus synthetic ops would stop paying that tax.

One last note: ITT's pin was a *branch* pin resolved at image-build time, so the Aug 6
merge silently changed what ITT shipped without any commit on the ITT side. Once ITT
moves to the release, this branch loses that downstream consumer and the incidental
test exposure that came with it.

PR #681, opened 2026-05-29
Fix eval harness scripts for standalone golden ref and repeatability runs

Ahmedhasssan-aig:
## Summary

Fixes shell and Python bugs that prevented `generate_ref.sh` and `run_repeatability_parallel.sh` from running in standalone mode on the eval node.

- **generate_ref.sh**: fix `$6` unbound variable, docker exec array expansion, tarball extraction path, agent failure detection, ref sanitization, and add `TEST_IDS` filter for single-case runs.
- **run_repeatability_parallel.sh**: define missing `TEST_TRACES_CSV` and pass empty `platform2` in standalone mode.
- **workflow_scripted_evals.py**: remove duplicate `_check_model_id` invocation so all workflow checks run correctly.


tsrikris (approved):
LGTM!

PR #685, opened 2026-05-29
Add graph capture gate for comparison extension

kyle-hoffmeyer:
Prevent users from calling tracediff_comparison_extension when passing graph capture folder as the extension uses TraceDiff which does not support graph capture.


devalshahamd (approved):
LGTM!

PR #687, opened 2026-05-30
Fix #603 adding too many events

kyle-hoffmeyer:
NOTE: Reference CSVs have been updated due to there being UID ordering change. Only UID values have changed from reference CSVs

Before PR https://github.com/AMD-AGI/TraceLens/pull/603, many events with malformed timestamps were discarded during tree building. PR https://github.com/AMD-AGI/TraceLens/pull/603 fixed this issue and allowed for these events that were previously discarded by Trace2Tree to be added back in. We call these events "bleeding" events. However, https://github.com/AMD-AGI/TraceLens/pull/603 introduced another bug which this PR aims to fix.

event here refers to the current event that Trace2Tree is trying to integrate into existing tree. The stack here refers to the current subtree that Trace2Tree is trying to integrate the event into.

Before https://github.com/AMD-AGI/TraceLens/pull/603, when event.start >= stack[-1].start and event.start <= stack[-1].end but event.end >= stack[-1].end, Trace2Tree would create a standalone tree with the event as the root. PR https://github.com/AMD-AGI/TraceLens/pull/603 changed this behavior such that the event does not become a standalone tree and is instead assigned as stack[-2]'s child. This is because there are many cuda runtime events where the timestamps are malformed and slightly overlaps with sibling events. These events are actually children of stack[-2], and we do not want to drop these.

However, there are some events with meaningless timestamp in the context of the call stack that were added into the subtrees through https://github.com/AMD-AGI/TraceLens/pull/603. There are pycapsule events with incredibly long runtimes that span much further than operations in the stack and should not be included in the subtree.

This PR introduces another guard for these types of functions. If there is a "bleeding" event, then it only attaches to stack[-2] if the event.end <= stack[-2].end and the event is a cuda_runtime, kernel, or cpu op. It does not look further beyond stack[-2] ([-3, -4, etc.]) on the basis that if event.end is longer than stack[-3].end or stack[-4].end then this event should not be a part of the call stack.


ajassani:
we need to understand why this regression was not caught in our CI

ajassani:
can we close this once we merge 694?

kyle-hoffmeyer:
Yes we can close this after 694

ajassani:
closing since 694 is merged

PR #689, opened 2026-06-01
Trace2Tree: guard graph launches against full-trace scans

ajassani:
## Summary
- Add a graph-launch regression test that fails if `_get_graph_gpu_events` reintroduces a per-launch full trace scan.
- Preserve existing CUDA graph linkage behavior while covering the prebuilt GPU-event index path.

## Test plan
- `python -m pytest tests/test_kernel_launchers.py -q`


gabeweisz (approved):

ajassani:
unnecessary

PR #694, opened 2026-06-02
Trace2Tree sub microsecond overlap tolerance

kyle-hoffmeyer:
**NOTE: Reference CSVs have been regenerated.**

Previously, if events overlap with their siblings, these events would be dropped or orphaned as part of their own subtree.
This PR adds these events to the subtree if the overlap is likely due to noise. By allowing for a 1 microsecond overlap, cuda_runtime events and kernels that were previously dropped from the subtree are now added back.
The events with more than 1 microsecond of overlap can be safely dropped without losing any kernels.

Across 20 traces, these are the number of events that are added back to subtree with this change:
| Bucket | Range | Count | % | Categories | Interpretation |
|--------|-------|-------|---|------------|----------------|
| Small | < 1 us | 1,597 | 80.7% | cuda_runtime, cpu_op, some python_function | Sub-microsecond. Likely profiling clock jitter / instrumentation overhead |
| Moderate | 1 us – 1 ms | 325 | 16.4% | python_function only | Real overlap. Event starts inside stack[-1] |
| Large | > 1 ms | 58 | 2.9% | python_function only | Massive overlap -> PyCapsule/built-in methods |

Reference CSVs have been updated since the "UID_first" values have been shifted by a constant value due to there being more events in the trace tree. The raw UID values come from the trace event metadata itself. However in the case of graph capture reports, when `merge_capture_trace_into_graph` merges the capture events into the graph tree, `update_subtree_uids_and_timestamps` regenerates all the UIDs of the capture trace event based on the number of capture trace events.

| Trace | main | fix/sub_microsecond_guard | Δ (s) | % |
|-------|-----:|-----:|-----:|----:|
| trace1 | 82.23 | 82.83 | +0.60 | +0.7% |
| trace2 | 0.42 | 0.43 | +0.01 | +2.3% |
| trace3 | 68.87 | 68.27 | -0.60 | -0.9% |
| trace4 | 71.71 | 71.41 | -0.30 | -0.4% |
| trace5 | 107.04 | 106.03 | -1.01 | -0.9% |
| trace6 | 1246.46 | 1247.37 | +0.91 | +0.1% |
| trace7 | 49.82 | 50.09 | +0.27 | +0.5% |
| trace8 | 228.86 | 227.73 | -1.13 | -0.5% |
| trace9 | 198.32 | 199.69 | +1.37 | +0.7% |


ajassani (approved):

ajassani:
Please merge this once Deval approves
Let me know if we need to approve again after merging main

devalshahamd (approved):
LGTM!

devalshahamd (approved):

PR #696, opened 2026-06-02
Add date and commit hash to generated wheel name

gabeweisz:
Add date and commit hash to generated wheel name


ajassani:
Lint test seems to fail

gabeweisz:
> Lint test seems to fail

Thanks - I'd run black but then did not realize that space was on purpose

ajassani (approved):

PR #707, opened 2026-06-03
Add graph capture + tracediff extension support

kyle-hoffmeyer:
## Summary

Extends the performance report pipeline to support running the TraceDiff comparison extension when **both** the baseline and comparison traces are graph capture traces. Previously, using `--capture_folder` with `--comparison_json_path` was explicitly blocked. This PR lifts that restriction with the `--comparison_capture_folder` flag.

---

### `generate_perf_report_pytorch_inference.py`

**New `--comparison_capture_folder` CLI argument**

Adds `--comparison_capture_folder`. When supplied, `merge_capture_trace_into_graph()` is called to augment the graph replay trace with the corresponding capture trace.

**Updated CLI validation**

The old guard that blocked `--capture_folder` + `--comparison_json_path` together is replaced. The new check requires that `--comparison_capture_folder` is always paired with `--comparison_json_path`.

---

### `TraceLens/Trace2Tree/trace_capture_merge_experimental.py`

**Regex hardening**

The `execute_context_..._generation_...` pattern now matches `[\d_]+` instead of `\d+` in the parenthesised groups, accommodating variants where the numeric fields contain underscores.

```
execute_context_3(512)_generation_2(256)      # matched before
execute_context_3(512_0)_generation_2(256_1)  # did NOT match — merge silently skipped
```

**Greedy alignment in `verify_subtree_events`**

Previously, a count mismatch between capture dispatch events and graph kernel events was an immediate failure. A new `_align_capture_to_graph()` helper performs a greedy forward scan: for each graph kernel, it scans forward through capture events until a dispatch with a matching kernel name is found, skipping unmatched dispatches. Note that Memcpy/Memset events are handled differently as these dispatchers don't have the kernel name in the args. If every graph kernel is matched, the surplus capture events are discarded and alignment succeeds. This makes the merge resilient to kernels that may have not been profiled or captured correctly.

**Cache key includes file path**

`_get_cached_capture_tree` previously keyed the LRU cache on the operation name string alone. This key conflicts when the baseline and comparison traces are passed through this function. The key is now `(name, abspath(filepath))`, preventing unintended cache hits.

---

## Usage

```bash
# Baseline + comparison, both graph capture traces
python generate_perf_report_pytorch_inference.py \
    --profile_json_path baseline.json \
    --capture_folder ./baseline_capture/ \
    --comparison_json_path comparison.json \
    --comparison_capture_folder ./comparison_capture/
```

---


TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -58,37 +58,96 @@ def get_subtree_events(tree, event, cat_filter=None, name_filter=None):
     return list_events, list_filtered_events
 
 
+def _capture_kernel_name(event):
+    return event.get("args", {}).get("kernel", event["name"])
+
+
+def _align_capture_to_graph(capture_events, graph_events):
```
devalshahamd, line 65: Why do we need this greedy alignment? to account for some dropped kernels? Maybe add that in the docstring "robust alignment path in case of minor mismatch due to profiler"

devalshahamd (commented):
One minor comment, otherwise looks good!

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
kyle-hoffmeyer: Yes, otherwise the entire alignment will fail. I've updated docstring.

devalshahamd (approved):

PR #720, opened 2026-06-05
exclude aiter test functions from shape profiling

### LINKED ISSUE #709: SGLang patch shape_discovery option resulting in error
When shape_discovery: True is set in the extra_body of its POST /start_profile request to an SGLang server running DeepSeek-V3 (TP=8) — and the server is also launched with --enable-shape-discovery-for-cuda-graph-profile — the kernel-shape-profiler is activated a second time during the live deferred profile capture window (the first activation already happened during cuda-graph capture/warmup). This second activation re-patches ~1923 op references across ~990 entry points while real traffic is in flight, including a step where prefill and decode batches are merged.

More details on the error:

Failure mode
On the very next prefill+decode mixed step, all 8 TP scheduler workers crash inside ScheduleBatch.merge_batch with:

RuntimeError: Expected all tensors to be on the same device, but got tensors is on cuda:N,
different from other tensors on cpu (when checking argument in method wrapper_CUDA_cat)
  File ".../sglang/srt/managers/schedule_batch.py", line 2510, in merge_batch
    self.seq_lens_cpu = torch.cat([self.seq_lens_cpu, other.seq_lens_cpu])

devalshahamd:
This pull request updates the `kernel_shape_profiler.py` patch for both `sglang_0_5_11` and `sglang_0_5_12` to improve the automatic discovery of relevant kernel modules and to avoid importing test, benchmark, and autotune modules that could interfere with profiling. The changes ensure that only the appropriate kernel modules are considered during profiling, reducing the risk of side effects from test or tuning modules.

Key improvements include:

**Module discovery improvements:**
* Updated the `_AUTO_DISCOVER_PREFIXES` tuple to specifically include `sglang.srt.layers.` and `sglang.srt.lora.` instead of the broader `sglang.srt.`, ensuring only relevant submodules are targeted for profiling.

**Import safety enhancements:**
* Added logic to skip importing modules whose names indicate they are test, benchmark, or autotune modules (e.g., names starting with `test_`, `bench_`, or ending with `_test`, `_tune`). This prevents side effects such as unintended changes to the device state that could corrupt profiling results.


mohbasit (commented):
Looks good, removed the reduction in scope for patches

mohbasit (approved):

tsrikris (approved):

PR #727, opened 2026-06-09
Roofline the Triton fused-MoE expert GEMM (invoke_fused_moe_kernel)

### LINKED ISSUE #726: Triton fused-MoE GEMM (invoke_fused_moe_kernel) falls into un-roofline'd `other` category → no optimization candidate
## Summary

The sglang/vLLM Triton fused-MoE expert GEMM — op `sglang_profiler::fused_moe_triton_kernels_invoke_fused_moe_kernel*` (GPU kernel `fused_moe_kernel`, parent `nn.Module: FusedMoE`) — is **correctly ranked** by TraceLens as the #1 GPU consumer, but the category taxonomy does **not** recognize it as a GEMM/grouped-GEMM nor as a MoE op, so it lands in the catch-all **`other`** category. `other` is handled by the generic-op analyzer, which emits **no roofline/efficiency metrics** (`efficiency_percent` null) and **empty** `category_findings` → **no `reasoning-candidate` block** is produced. Net: the single largest editable kernel yields no optimization candidate, so a downstream kernel-tuning consumer never gets a target from TraceLens.

## Affected component (exact location)

- **Repo:** `AMD-AGI/TraceLens` (observed on `main` @ `559cda22`)
- **Categorizer:** `TraceLens/PerfModel/torch_op_mapping.py` — `categorize_torch_op()` / `_categorize_torch_op_from_registry()` and `resolve_perf_model_class()`. The op is in neither `op_to_perf_model_class_map`, `CATEGORY_ONLY_OP_MAPPING`, nor `OP_CATEGORY_PATTERNS` (`gemm` only catches `aten::mm`-family Tensile calls; `MoE_aux` only catches `aiter::*` routing ops), and there is no perf-model matcher for the Triton `fused_moe_kernel` → categorizes as `other`, `has_perf_model = False`.
- **Manifest / findings:** `TraceLens/Agent/Analysis/utils/orchestrator_prepare.py` buckets by `op category` (→ `other`), and `category_analyses/other_analysis.py` (generic-op analyzer) is non-roofline → empty `category_findings`.

## Root cause

There is no perf model nor roofline-capable category for the Triton `fused_moe_kernel`. Because `op category = other` and no perf model resolves, the perf report leaves `GFLOPS`, `Data Moved (MB)`, `FLOPS/Byte`, `Pct Roofline`, `Compute Spec` empty for the op, and the orchestrator routes it to the generic-op analyzer which (by design) produces no roofline efficiency and no kernel-tuning impact estimate.

## Impact (real, observed)

In a live run of our internal kernel-optimization pipeline on **Qwen3-30B-A3B (sglang, bf16, MI300X, Triton MoE backend)**:

- `fused_moe_kernel` = **321 ms = 67%** of the analyzed chunk (the single largest, fully editable kernel), correctly surfaced in `ops_summary.csv` — but with category `['other']`.
- `category_manifest.json`: `other` is the **largest** category (~377 ms) yet routed to the generic-op analyzer.
- `category_findings/other_findings.md`: **empty** → no `reasoning-candidate` block for the dominant kernel.
- Effect: TraceLens drops the 67% editable kernel as an optimization target, so the kernel-tuning consumer never receives a roofline-grounded candidate for it.

## Evidence

Run dir `…/Qwen-Qwen3-30B-A3B/20260609T190334Z/kernel-agent/runs/.../tl-4b9bac13/tracelens/`:

- `perf_report_csvs/ops_summary.csv`: `sglang_profiler::fused_moe_triton_kernels_invoke_fused_moe_kernel_427`, `parent_module = nn.Module: FusedMoE`, `total_direct_kernel_time_ms = 321.06`, `Percentage (%) = 67.16`, `Categories = ['other']`.
- `perf_report_csvs/unified_perf_summary.csv`: two rows for the op (gate/up + down GEMMs) — `op category = other`, `has_perf_model = False`, and empty `GFLOPS` / `Data Moved (MB)` / `FLOPS/Byte` / `Pct Roofline` / `Compute Spec`.
- `category_data/category_manifest.json`: `other` is the largest category.
- `category_findings/other_findings.md`: empty findings.

The op is a *grouped* GEMM over experts and is roofline-able from the kernel args: gate/up `[[T,K],[E,N,K],…,[T*topk,N],…]` and down `[[T*topk,K],[E,N,K],…,[T,topk,N],…]`. Modelled as `FLOPs = 2·M·K·N` (M = `num_tokens × top_k`), the two GEMMs run at ~17% / ~10% of the bf16 matrix roofline — a strong, compute-bound optimization candidate that is currently invisible.

## Suggested fix

Give the Triton fused-MoE expert GEMM a roofline-capable category + perf model, matched by **op/kernel-name pattern** (`invoke_fused_moe_kernel`) rather than the single instance (the registered op name carries an unstable numeric suffix and the W4A16/gptq-awq path reuses the same kernel):

- Add a grouped-GEMM roofline perf model for `*invoke_fused_moe_kernel` (FLOPs `2·M·K·N`, `M = num_tokens × top_k`, `K` from the activation so packed W4A16/FP8 weights are handled), reusing the existing `GroupedGemm` machinery.
- Categorize it under an existing roofline-capable, fully-wired MoE category (`MoE_fused` → moe-analyzer) so it emits `efficiency_percent` + a `reasoning-candidate` block.

A fix PR follows.

Umangatamd:
## Summary

The sglang/vLLM Triton fused-MoE expert GEMM — op `..._invoke_fused_moe_kernel` (GPU kernel `fused_moe_kernel`, parent `nn.Module: FusedMoE`) — was correctly **ranked** as the #1 GPU consumer but categorized as the catch-all **`other`**, which is handled by the non-roofline generic-op analyzer: `efficiency_percent` was null and `category_findings` empty, so **no `reasoning-candidate`** was produced for the single largest editable kernel (~67% of GPU time on Qwen3-30B-A3B). This PR gives it a roofline-capable category + perf model so it emits efficiency metrics and an optimization candidate.

## Changes

- **`moe_triton_fused_gemm`** (`PerfModel/extensions/moe_perf_model_extensions.py`): a `GroupedGemm` subclass (`category = "MoE_fused"`) modelling **one** expert-GEMM launch (gate/up or down). `FLOPs = 2·M·K·N` where `M` = routed token-expert rows (`num_tokens × top_k`, read from the output operand), `K` is taken from the **activation** (so storage-packed W4A16/FP8 weights still yield the logical contraction dim), and `N`/`E` from the expert-weight tensor. Shapes are read generically (first 3-D operand = weights; output = non-weight operand whose last dim is `N`) so both projections route here regardless of their differing layouts.
- **Perf-model matcher** (`PerfModel/torch_op_mapping.py`): matches the stable `invoke_fused_moe_kernel` substring (the registered op name carries an unstable numeric suffix, and the W4A16 gptq/awq path reuses the same kernel), mirroring the existing triton-compiled matcher.
- **Categorization fallback** (`_categorize_torch_op_from_registry`): pattern-matched perf models inherit their declared `category` as the final fallback, keeping categorization and roofline in sync. No behavior change for existing ops (they are resolved earlier by the registry / `OP_CATEGORY_PATTERNS` / kernel-name fallback).

After the fix the op categorizes as `MoE_fused` (→ moe-analyzer, fully wired) and the two expert GEMMs report **~17%** (gate/up) and **~10%** (down) of the bf16 matrix roofline — compute-bound, low-efficiency → a `reasoning-candidate` is emitted.

## Files changed

- `TraceLens/PerfModel/extensions/moe_perf_model_extensions.py` — new `moe_triton_fused_gemm`
- `TraceLens/PerfModel/extensions/__init__.py` — export it
- `TraceLens/PerfModel/torch_op_mapping.py` — matcher + perf-model-category fallback
- `tests/test_fused_moe_triton_gemm.py` — new unit tests

## Test plan

- [x] `pytest tests/test_fused_moe_triton_gemm.py` (10 passed): suffix-agnostic resolution + `MoE_fused` categorization across name variants, FLOPs/bytes for gate/up (2-D output) and down (3-D output) using real Qwen3-30B-A3B shapes, logical `K` from activation for packed weights, backward unsupported, and missing-weight error.
- [x] No regression: `pytest tests/test_torch_op_categorization_registry.py tests/test_primus_turbo_grouped_gemm.py tests/test_primus_op_categorization.py tests/test_analysis_agent.py` (and others) pass; unrelated ops keep their categories (`aten::mm`→`GEMM`, unknown→`other`). The single pre-existing failure `test_roofline_bound.py::test_origami_time_in_unified_perf_summary` is environmental (the `origami` package is not installed) and also fails on clean `main`.

Closes #726


devalshahamd:
Hi @Umangatamd 

A few things:
1) We are doing the mapping for all inference related ops in https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/extensions/pseudo_ops_perf_utils.py and not the main torch_op_mapping.
2) Can you provide me the link to this op source code? We can see if we can reuse the FusedMoE base class. We need to account for activated experts, packed weight etc for MoE ops' roofline, so it is better to classify them as FusedMoE  and not grouped  GEMM.

tsrikris:
@Umangatamd Do you mind sharing traces? Might be easier for us to implement this with that in hand. Additionally, can you point us to the source code?

Umangatamd:
Extended this PR with the **shape-capture** half of the fused-MoE gap (commit dea1aec5, Closes #731).

The roofline-unresolved surfacing already in this PR makes the dominant fused-MoE kernel reach the report, but its `Args`/shapes were still empty because the kernel arrives as a dimensionless pybind built-in. `build_operation_metrics` now recovers the operand shapes from `perf_report_csvs/ops_unique_args.csv` (the gate/up + down grouped-GEMM operands, keyed by the embedded `invoke_fused_moe_kernel` symbol) and renders them via the existing `format_args`, scoped to that op so nothing else regresses.

Validated on a real Qwen3-30B-A3B MoE decode trace — recovered operands match a hand-written GEAK harness's shapes:
- gate/up: `(15360,2048)`, `(128,1536,2048)`, `(122880,1536)` bf16
- down: `(122880,768)`, `(128,2048,768)`, `(15360,8,2048)` bf16

A sibling fix in the internal kernel-optimization pipeline back-fills the same operand shapes onto the emitted kernel candidate so its shape-gated kernel-opt dispatch passes with `shape_provenance=torch_trace` (the rendered report is not relied on as a deterministic shape carrier).

tsrikris:
@Umangatamd  we are going to be rejecting this PR. Pls file a ticket with a trace the kernel you'd like a perf model for, the TraceLens team will handle it.

Umangatamd:
Thanks — clarifications, one concession, and the trace/source you asked for:

- **Source:** `invoke_fused_moe_kernel` is the Triton fused-MoE expert GEMM in the sglang/vLLM fused-MoE Triton path (`.../layers/moe/fused_moe_triton/fused_moe.py`); the GPU kernel is `fused_moe_kernel`. I'll add the exact upstream permalink in-thread.
- **Category:** it already emits `category = "MoE_fused"` (line 1628), so it routes to the MoE analyzer, not GroupedGEMM. The only open question is the *base class*.
- **Base class:** I subclassed `GroupedGemm` because this kernel is launched once per projection (gate/up, then down) — each launch is a single grouped GEMM (`2*M*K*N`, `M` = actual routed token-expert rows). The `FusedMoE` base models the whole MoE (FC1+FC2+gated+aggregation) in one shot, which would double-count when applied to a single-projection kernel. If you'd prefer it under `FusedMoE` for code reuse, I think we'd need a per-projection variant of that base.
- **Packed weight:** already handled — `K` is taken from the activation's trailing dim, not the storage-packed expert weight, so W4A16/FP8 weights don't corrupt `K` (lines 1657–1663).
- **Activated experts:** good catch on the memory side — `GroupedGemm.bytes` counts all `G` experts' weights, so the *bytes/AI* figure overcounts weight traffic at small batch. The *compute* roofline (which is what surfaces the candidate here, and is compute-bound) uses actual routed `M` and is correct. Happy to refine `bytes()` to use an active-expert estimate (mirroring `FusedMoE.bytes_func`'s `E_active`) so the memory roofline is accurate too.
- **Placement:** the perf-model class lives in `moe_perf_model_extensions.py` with the other MoE models; I registered it via a substring matcher (`torch_op_mapping.py`) rather than `get_pseudo_op_mappings()` because the op name carries an unstable numeric suffix (e.g. `..._invoke_fused_moe_kernel_427`) and the W4A16 gptq/awq path reuses the same kernel, so an exact-name entry won't match. If you'd prefer it under the pseudo-op mapping, I can add prefix/substring support there and move it.
- **Trace to reproduce:** happy to share the representative trace + the exact expert-GEMM operand shapes with you directly / on the internal tracker so you can reproduce the roofline.

PR #742, opened 2026-06-16
LayerNormLinear perf model

kyle-hoffmeyer:
# Perf models for _LayerNormLinear and _LayerNormLinearBackward

`_LayerNormLinear` is a Transformer Engine v2 op that fuses LayerNorm and a Linear transformation into a single CPU-side op. `_LayerNormLinearBackward` is its autograd backward.

The linear transformation is handled by the `tev2_pseudo_gemm` perf model via pseudo op injection, which reparents the GEMM kernel out from under `_LayerNormLinear` into its own synthetic host op (`_LayerNormLinear_yfwd_mm`). This PR covers only the LayerNorm portion that remains under the original op after the GEMM is extracted.

The new perf models for these ops have been folded into the existing `te_layer_norm_fwd` and `te_layer_norm_bwd` perf models.

---

### `_LayerNormLinear.forward()` ([layernorm_linear.py](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/module/layernorm_linear.py))

```python
@staticmethod
def forward(
    ctx,
    inp: torch.Tensor,              # [0] input activation, e.g. (8192, 4, 4096)
    ln_weight: torch.Tensor,        # [1] LayerNorm gamma, e.g. (4096,)
    ln_bias: Union[torch.Tensor, None],  # [2] LayerNorm beta, e.g. (4096,) or empty
    weight: torch.Tensor,           # [3] Linear weight, e.g. (28672, 4096)
    weight_workspace: Optional[torch.Tensor],  # [4] FP8 workspace
    bias: torch.Tensor,             # [5] Linear bias
    non_tensor_args: Tuple,          # [6] config tuple (eps, fp8, ...)
) -> Tuple[torch.Tensor, ...]:
```

PyTorch profiler records each arg's shape as `Input Dims[i]` in this order.

---

## Existing `te_layer_norm_fwd` Validation

```python
# examples/example_megatron_extension.py:670
def get_param_details(event):
    args_input_dims = event["args"]["Input Dims"]
    op_shape = tuple(args_input_dims[0])
    weight_shape = args_input_dims[1]
    num_channels = weight_shape[0] if weight_shape else op_shape[-1]
    has_bias = bool(args_input_dims[2])
    ...
```

### Arg-to-parameter mapping

| Expression | Value | Why |
|------------|-------|-----|
| `Input Dims[0]` → `op_shape` | `(8192, 4, 4096)` | Index 0 = `inp` in TE's forward signature. This is the input to `nvte_layernorm_fwd(x, ...)`. |
| `Input Dims[1]` → `weight_shape` | `(4096,)` | Index 1 = `ln_weight` = `gamma`. Shape `[H]`. |
| `weight_shape[0]` → `num_channels` | `4096` | First dim of `gamma[H]` gives `H`, the normalization dimension. |
| `bool(Input Dims[2])` → `has_bias` | `True/False` | Index 2 = `ln_bias` = `beta`. Empty when `ln_bias=None`. |
| `Input type[0]` → `dtype_in` | `"c10::bfloat16"` | Dtype of `inp`. |

### Bytes

Inherited from `Normalization.bytes_func` with `is_training=True`, `is_affine=True`:
bpe_in = bytes_per_element_input
bpe_out = bytes_per_element_output

```python
# perf_model.py:4370
activation_bytes = num_elems * bpe_in + num_elems * bpe_out   # read x + write z
weight_bytes = num_weight_tensors * num_channels * bpe_in     # gamma, beta, stats
```

| Component | Model | TE kernel |
|-----------|-------|-----------|
| Read `x[M,H]` + write `z[M,H]` | `2 * M*H * bpe` | `2 * M*H * bpe` |
| Read `gamma[H]`, `beta[H]` | `2 * H * bpe` | `2 * H * bpe` |
| Write `mu[M]`, `rsigma[M]` | modeled as `2 * H * bpe` | actually `2 * M * 4` (fp32, per-row) |

> **Note:** The calculation of `weight_bytes` is slightly inaccurate. This is an artifact of the `Normalization` base class being designed around BatchNorm, where statistics are per-channel (`[C]`), whereas LayerNorm statistics (`mu`, `rsigma`) are per-token (`[M]`, fp32). The fix would be to add `M * 4 * 2` bytes for `mu` and `rsigma` directly instead of folding them into `num_weight_tensors * num_channels * bpe_in`. The same inaccuracy applies to all LayerNorm perf models that inherit `bytes_func`. A follow-up PR should fix `Normalization.bytes_func` and `bytes_func_bwd` to handle the LayerNorm case correctly.

The existing `te_layer_norm_fwd` works fine for `_LayerNormLinear`. A new perf model is not necessary.

---

### The shape problem

The backward data flow through the fused op is:

```
dY [M, H_out]  →  dgrad GEMM (@ W)  →  dX_linear [M, H_in]  →  LayerNorm bwd  →  dx [M, H_in]
```

The GEMM backward (dgrad + wgrad) is already handled by the `tev2_pseudo_gemm` pseudo op perf model. The `te_layer_norm_bwd` perf model only needs to cover the LayerNorm backward, which operates on `dX_linear [M, H_in]`.

The problem is that the profiler records `Input Dims[0]` as `dY [M, H_out]`, the gradient flowing in from the loss before the dgrad GEMM runs. However, the LayerNorm backward kernel receives `dX_linear [M, H_in]` after the GEMM has projected the gradient back to the hidden dimension. Since `dX_linear` is an intermediate tensor computed inside the fused op, the profiler has no record of it.

`_annotate_layernorm_backward_shapes()` fixes this by copying from the forward:

```python
# examples/example_megatron_extension.py:91
ln_input_shape = fwd_event["args"]["Input Dims"][0]   # (8192, 4, 4096)
ln_weight_shape = fwd_event["args"]["Input Dims"][1]   # (4096,)
bwd_op["args"]["_ln_input_shape"] = ln_input_shape
bwd_op["args"]["_ln_weight_shape"] = ln_weight_shape
```

The perf model uses the annotated shapes:

```python
# examples/example_megatron_extension.py:713
def get_param_details(event):
    args = event["args"]
    if "_ln_input_shape" not in args:
        raise ValueError("missing '_ln_input_shape' annotation")
    op_shape = tuple(args["_ln_input_shape"])           # (8192, 4, 4096) — H_in
    num_channels = args["_ln_weight_shape"][0]           # 4096 — H_in
    ...
```

| Expression | Value | Why |
|------------|-------|-----|
| `_ln_input_shape` → `op_shape` | `(8192, 4, 4096)` | Copied from forward's `Input Dims[0]` = `inp`. Same tensor shape that `nvte_layernorm_bwd(dz, x, ...)` operates on. |
| `_ln_weight_shape[0]` → `num_channels` | `4096` | Copied from forward's `Input Dims[1]` = `ln_weight[H]`. Matches `nvte_layernorm_bwd`'s `gamma[H]`. |
| `has_bias=False` | — | TE backward doesn't read `beta`; `dbeta = sum(dz)` needs no bias tensor read. |
| `is_affine=True` | — | `gamma` is always present; backward reads it to compute `dy = gamma * dz`. |

### Bytes

Inherited from `Normalization.bytes_func_bwd` with `output_mask=[True, True, False]`:

```python
# perf_model.py:4450
bytes  = num_elems * bpe_out                        # read dz (output of forward stage)
bytes += num_elems * bpe_in * 2                     # read x (input of forward stage) + write dx (training, mask[0])
bytes += num_channels * bpe_in                      # read rsigma (training)
bytes += num_channels * bpe_in                      # read gamma (affine)
bytes += num_channels * bpe_in                      # write dgamma (affine, training)
```

| Component | Model | TE kernel |
|-----------|-------|-----------|
| Read `dz[M,H]` + read `x[M,H]` + write `dx[M,H]` | `3 * M*H * bpe` | `3 * M*H * bpe` |
| Read `gamma[H]`, write `dgamma[H]` | `2 * H * bpe` | `2 * H * bpe` |
| Read `mu[M]`, `rsigma[M]` | modeled as `1 * H * bpe` | actually `2 * M * 4` (fp32, per-row) |
| Write `dbeta[H]` | not counted (`has_bias=False`) | `1 * H * bpe` |

Activation bytes (dominant term) match exactly. Same minor discrepancy as forward.

---


devalshahamd (approved):
LGTM!

PR #743, opened 2026-06-16
Enhance call stack info in TraceLens report

kyle-hoffmeyer:
## Overview

Adds full, structured call stacks and an `entry_point` column to the `unified_perf_summary` sheet when `--include_call_stack` is passed to `TraceLens_generate_perf_report_pytorch` (and its inference variant).

`unified_perf_summary` gains two columns:

| Column | Type | Description |
|---|---|---|
| `entry_point` | str | First `.py` frame containing the op name. |
| `call_stack_full` | str (list repr) | Full call chain from process root to GPU kernel(s). |

`unified_perf_callstacks.csv` is deleted.

Previously, `--include_call_stack` produced a truncated string column (`trunc_call_stack`) and a separate `unified_perf_callstacks` sheet built by re-walking the event tree via `ex_UID` at report-generation time. The new approach propagates call stacks incrementally during tree traversal, so each kernel already carries its full ancestry when the summary is assembled. This is more accurate (the stack reflects the exact path each GPU kernel was dispatched from, not just the path to its parent op), more efficient (no second tree walk), and produces a richer output.

### Phase 1: top-down call stack accumulation (`traverse()`)

The first phase builds the call stack from the **process root down to the collected CPU op**. This happens inside `traverse()` in `_collect_events_with_gpu_work`, which recursively descends the CPU op tree to find ops worth reporting (ops with perf models, leaf kernel launchers, etc.).

As `traverse()` descends, it accumulates a `call_stack` list — appending filtered frame names (`nn.Module`, `::`, `/`) at each level. When an op is collected, the accumulated stack is stored on it as `event["_call_stack"]`.

When `add_python_func=True`, traversal starts from parentless `python_function` roots (before falling through to `cpu_root_nodes`), ensuring Python-level frames (file paths, module names) that sit above the CPU op tree are captured.

### Phase 2: bottom-up per-kernel suffix (`_get_op_row_data`)

Phase 1 gives us the stack from root to the collected CPU op (e.g. `aten::mm`). Phase 2 recovers call stack between the kernel and the CPU op by walking **bottom-up** from each GPU kernel (via `get_parent_event`) back to the collected op, collecting matching frames along the way.

**Why two phases instead of one?** `traverse()` operates on the CPU op tree via `children` links. GPU kernels are not children in this tree — they are linked through the separate `gpu_events` side-channel and live on a different pid/tid. `traverse()` has no mechanism to descend into GPU events. The per-kernel suffix walk in `_get_op_row_data` is the only place where we iterate `gpu_events` and access their parent chains, so it's the natural place to capture the remaining frames.

### Assembly: common prefix + diverging tails

In `get_unified_perf_summary`, the per-kernel call stacks are combined into a single `call_stack_full` value per summary row:

- **Single-kernel ops**: flat list from root to kernel.
- **Multi-kernel ops**: longest common prefix as flat elements, then each unique diverging tail as its own sub-list. This avoids repeating the shared ancestry while clearly showing where kernel paths diverge.

## Entry point extraction

`_find_entry_point` identifies the Python function that dispatches each CPU op. It uses a two-phase strategy, both starting from the CPU op's position in the flattened call stack:

### Phase 1: Inward matching

From the CPU op, search **downward** (toward leaf/kernel) for a `.py` frame whose function name contains the op's local name (the part after `::`). This catches cases where a Python function with a matching name sits below the op — e.g. `vllm::unified_attention_with_output` finds `vllm/attention/layer.py(858): unified_attention_with_output` as a child frame.

### Phase 2: Outward matching (fallback)

If inward matching finds no match, walk **upward** from the CPU op (toward root), skipping wrapper functions, and return the first non-wrapper `.py` frame. This handles generic `aten::` ops (e.g. `aten::copy_`, `aten::mm`, `aten::layer_norm`) that have no matching Python function name below them.

## Output format

With `--include_call_stack`, `unified_perf_summary` gains two columns:

| Column | Type | Description |
|---|---|---|
| `entry_point` | str | First `.py` frame containing the op name. |
| `call_stack_full` | str (list repr) | Full call chain from process root to GPU kernel(s). |

#### Debug mode

Set `TRACELENS_DEBUG=1` to add three diagnostic columns to `unified_perf_summary`:

| Column | Description |
|---|---|
| `num_wrappers` | Number of call stack frames between the entry point and the CPU op |
| `traversal` | Which strategy matched: `inward` or `outward` |
| `wrappers` | The full frame list from entry point to CPU op (inclusive) |

## Example:

`call_stack_full` represents call stack from root to kernel(s) via nested lists.

Single divergence:

['csrc/cpp_itfs/pa/pa_ragged.py(65): paged_attention_ragged',
 ['void paged_attention_ll4mi_QKV_mfma16_kernel<...>'],
 ['void paged_attention_ll4mi_reduce_kernel<...>']]

Here, `paged_attention_ragged` calls two kernels and diverges.

Nested divergence:

['aiter/ops/fake_multi_op.py(100): fake_multi_launcher',
 ['aiter/ops/sub_path.py(200): sub_launcher', 'fake_kernel_e<...>'],
 ['aiter/ops/sub_path.py(200): sub_launcher', 'fake_kernel_f<...>'],
 ['fake_kernel_d<...>']]

Here, `fake_multi_launcher` calls `fake_kernel_d` and `sub_launcher` and diverges. But within `sub_launcher`, it diverges again. Note that nested divergences are not double-nested and each unique end-path gets its own flat sublist, with shared sub-prefixes repeated.

Number of events between entry point and CPU op (lower is better):
image

## Runtime impact
No impact on runtime

### Without `--include-call-stack`

| Trace | Baseline (s) | Current (s) | Delta (s) | Delta (%) |
|---|---|---|---|---|
| facebook_timesformer (small) | 0.46 | 0.46 | +0.00 | +0.3% |
| gptoss_vllm_example0 (medium) | 74.48 | 73.68 | -0.80 | -1.1% |
| huvideo_rank4_step3 (large) | 114.67 | 111.67 | -3.00 | -2.6% |

### With`--include-call-stack`
| Trace | Baseline (s) | Current (s) | Delta (s) | Delta (%) |
|---|---|---|---|---|
| facebook_timesformer (small) | 0.46 | 0.46 | +0.00 | +0.2% |
| gptoss_vllm_example0 (medium) | 74.37 | 75.41 | +1.04 | +1.4% |
| huvideo_rank4_step3 (large) | 113.62 | 111.36 | -2.26 | -2.0% |


TraceLens/Reporting/generate_perf_report_pytorch.py
```diff
@@ -25,6 +27,151 @@
...
+    "triton/backends/",
+)
+
+_WRAPPER_NAME_PREFIXES = (
+    "<built-in",
+    "pybind11_builtins",
+    "nn.Module:",
+)
+
+_WRAPPER_FUNC_NAMES = frozenset({
```
gabeweisz: Any reason not to make them all frozenset?

gabeweisz (changes requested):
Needs a test

TraceLens/Reporting/generate_perf_report_pytorch.py
gabeweisz: Looks like there are plenty of tests that cover this, so you just need to check and fix the reference results and address my question about making them all frozenset (or none of them, the main thing is that this should be consistent or we need a reason why)

kyle-hoffmeyer: They can all be frozenset. Fixed

gabeweisz (approved):

kyle-hoffmeyer:
> Needs a test

Added test and hooked up into regression-tests.yml

ajassani (approved):

devalshahamd (approved):
LGTM!

PR #754, opened 2026-06-19
Fix sed pattern

kyle-hoffmeyer:
The magpie benchmark profiling skill tries to edit the `num_prompts=$((max_concurrency * 1))` string in benchmark_lib.sh in InferenceX. But this string is actually `num_prompts=$max_concurrency`: https://github.com/SemiAnalysisAI/InferenceX/blob/1ae2b9c6cfc022d77a13463fbdebea5f4acc7a98/benchmarks/benchmark_lib.sh#L355


devalshahamd (approved):
LGTM!

tsrikris (approved):

PR #758, opened 2026-06-23
Read nn.module names for model identification

kyle-hoffmeyer:
This PR makes model identification more robust.
Previously, when determining the model that a trace was ran on, the agent would only look at operation names and input dimensions for operations. However, the agent would occasionally misclassify traces if two models have a very similar architecture (Kimi-K2 and DeepseekvV3). 

Now, the agent looks at the names of the nn.Modules as well (when available).

The agent now correctly classifies Kimi-K2 traces. Verified that the agent still correctly classifies all the traces in the standalone tests.

### DIFF
--- a/TraceLens/Agent/Analysis/.cursor/agents/model-identification-agent.md
+++ b/TraceLens/Agent/Analysis/.cursor/agents/model-identification-agent.md
@@ -56,19 +56,24 @@ Execute the Python script to extract the **name**, **Input type**, and **Input D
 ```bash
 <prefix> python3 -c "
 import sys
-from TraceLens.Agent.Analysis.utils.report_utils import extract_condensed_op_info
-if not extract_condensed_op_info('<output_dir>', '<comparison_scope>'):
+from TraceLens.Agent.Analysis.utils.report_utils import prepare_model_identification_data
+if not prepare_model_identification_data('<output_dir>', '<comparison_scope>'):
     sys.exit(1)
 "
 ```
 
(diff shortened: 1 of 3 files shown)


kyle-hoffmeyer:
@tsrikris, the Kimi-K2 traces actually have nn.Module calls with KimiK2 in the name in addition to Deepseek. The issue seemed to be that the agent was only looking at operation dimensions. When the agent saw the nn.Module names, it knew that KimiK2 is built off of Deepseek architecture and that the Deepseek nn.Module calls are expected.

tsrikris:
PR looks good! Do you mind sharing a couple of sample traces to make sure it's functioning as designed?

kyle-hoffmeyer:
`model_info.json` outputs (what's used to populate appendix in report):

Kimi-K2 trace:

```
{
  "model": "Kimi K2 (DeepSeek-V3 architecture)",
  "architecture": "Transformer (MLA + MoE)",
  "scale": "671B MoE (37B active)",
  "precision": "BF16 (MoE weights: MXFP4, KV cache: FP8)"
}
```

Llama 3.2 trace (this doesn't have nn.Module calls; still populates correctly): 

```
{
    "model": "Llama 3.2 Vision",
    "architecture": "Transformer",
    "scale": "11B",
    "precision": "BF16"
}
```

Timesformer trace:

```
{
    "model": "TimeSformer",
    "architecture": "Transformer",
    "scale": "base",
    "precision": "BF16"
}
```

tsrikris:
LGTM!

PR #764, opened 2026-06-25
fix: add fallback for FP8 dtype-to-bytes mapping in JAX TreePerf

brieflynn:
# PR: Add fallback for FP8/S8 dtypes in `dtype_to_bytes` lookup

## Summary

Replace the direct `dtype_to_bytes[dtype]` lookup with `dtype_to_bytes.get(dtype)` and an inline fallback. Dtypes starting with `f8` or `s8` map to 1 byte. Truly unknown dtypes log a warning and the operand is skipped.

**File:** `TraceLens/TreePerf/tree_perf.py`, around line 3249

Before:
```python
if shape and dtype:
    total_input_bytes = (
        total_input_bytes + np.prod(shape) * dtype_to_bytes[dtype]
    )
```

After:
```python
if shape and dtype:
    nbytes = dtype_to_bytes.get(dtype) or (
        1 if dtype.startswith(("f8", "s8")) else None
    )
    if nbytes is None:
        logger.warning(
            "Unknown dtype '%s' in operand, skipping byte count", dtype
        )
        continue
    total_input_bytes = total_input_bytes + np.prod(shape) * nbytes
```

## Testing

- Ran report generation against an FP8 Mixtral trace that was previously crashing with a `KeyError`. Report generated successfully with correct byte counts.
- All 13 regression tests in the suite passed.

# Pull Request Template

> **Note to AMDers:**  

### DIFF
--- a/TraceLens/TreePerf/tree_perf.py
+++ b/TraceLens/TreePerf/tree_perf.py
@@ -3247,9 +3247,15 @@ class JaxTreePerfAnalyzer(TreePerfAnalyzer):
             for operand in operands:
                 dtype, shape, layout = parse_dtype_shape_layout(operand)
                 if shape and dtype:
-                    total_input_bytes = (
-                        total_input_bytes + np.prod(shape) * dtype_to_bytes[dtype]
+                    nbytes = dtype_to_bytes.get(dtype) or (
+                        1 if dtype.startswith(("f8", "s8")) else None
                     )
+                    if nbytes is None:
+                        logger.warning(
+                            "Unknown dtype '%s' in operand, skipping byte count", dtype
+                        )
+                        continue
+                    total_input_bytes = total_input_bytes + np.prod(shape) * nbytes
 
             total_input_bytes_list.append(total_input_bytes)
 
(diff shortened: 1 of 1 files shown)


TraceLens/TreePerf/tree_perf.py
```diff
@@ -3444,9 +3247,15 @@ def parse_dtype_shape_layout(operand):
             for operand in operands:
                 dtype, shape, layout = parse_dtype_shape_layout(operand)
                 if shape and dtype:
-                    total_input_bytes = (
-                        total_input_bytes + np.prod(shape) * dtype_to_bytes[dtype]
+                    nbytes = dtype_to_bytes.get(dtype) or (
```
gabeweisz: I would use the default parameter of get instead of "or" - make this
 nbytes = dtype_to_bytes.get(dtype, 1 if dtype.startswith(("f8", "s8")) else None)

gabeweisz (changes requested):

gabeweisz (approved):

PR #766, opened 2026-06-25
fix: resolve relative path in JAX protobuf file processing

brieflynn:
# PR: Fix relative path resolution for protobuf sidecar file lookup

## Summary

Wrap `protobuf_file_name` with `os.path.abspath` before calling `os.path.dirname` in `JaxProfileProcessor.process_protobuf_file`.

**File:** `TraceLens/util.py`, line 108

Before:
```python
dir_name = os.path.dirname(protobuf_file_name) + "/"
```

After:
```python
dir_name = os.path.dirname(os.path.abspath(protobuf_file_name)) + "/"
```

`os.path.abspath` normalizes a bare filename to an absolute path, so `os.path.dirname` returns the correct directory. This ensures `glob.glob` searches the right location for HLO `.pb` sidecar files.

## Testing

- Ran the fix against an FP8 Mixtral trace that was previously failing when invoked with a relative path. Report generated successfully.
- All 13 regression tests in the suite passed.

# Pull Request Template

> **Note to AMDers:**  

### DIFF
--- a/TraceLens/util.py
+++ b/TraceLens/util.py
@@ -105,7 +105,7 @@ class JaxProfileProcessor:
                 raw_to_tool_data as convert,
             )
 
-        dir_name = os.path.dirname(protobuf_file_name) + "/"
+        dir_name = os.path.dirname(os.path.abspath(protobuf_file_name)) + "/"
         hlo_filename = glob.glob(dir_name + os.path.sep + module_name + "*hlo_proto.pb")
         if len(hlo_filename) != 1:
             convert.xspace_to_tool_names([protobuf_file_name])


gabeweisz (approved):

TraceLens/util.py
```diff
@@ -105,7 +105,7 @@ def process_protobuf_file(protobuf_file_name, module_name):
                 raw_to_tool_data as convert,
             )
 
-        dir_name = os.path.dirname(protobuf_file_name) + "/"
+        dir_name = os.path.dirname(os.path.abspath(protobuf_file_name)) + "/"
         hlo_filename = glob.glob(dir_name + os.path.sep + module_name + "*hlo_proto.pb")
```
gabeweisz, line 109: If you are touching this code it would be nice to make this bit a "path.combine" instead of dir_name + os.path.sep + module_name + "*hlo_proto.pb" but that is not directly related to your PR and can be saved for another time if you are not making more commits

gabeweisz (approved):

PR #767, opened 2026-06-25
Feat: Perf Models for DSR1 MOE Routing: topk and scatter-gather  

spandoesai:
Adds TraceLens perf models for two MoE auxiliary kernels that were previously only category-tagged (MoE_aux)

1. BiasedGroupedTopk — models aiter::biased_grouped_topk_hip, the DeepSeek-style grouped router top-k selection (sigmoid + bias + grouped partial sort). Note that this model adds FLOPs for each arithmetic operation performed, might need review.

2. MoeSortScatterGather — models aiter::moe_sorting_fwd and aiter::mxfp4_moe_sort_hip, the token sort/scatter (permute) kernels that lay tokens out contiguously per expert.


TraceLens/PerfModel/extensions/moe_perf_model_extensions.py
```diff
@@ -1714,3 +1716,253 @@ def get_compute_precision(self):
...
+            "topk": topk,
+            "num_expert_group": max(1, num_expert_group),
+            "topk_group": max(1, topk_group),
+            "logits_dtype": types[0] if len(types) > 0 else None,
+            "bias_dtype": types[1] if len(types) > 1 else None,
+            "weights_dtype": types[2] if len(types) > 2 else None,
+            "ids_dtype": types[3] if len(types) > 3 else None,
+        }
+
+    def flops(self):
```
devalshahamd, line 1841: This is a memory bound op. The compute flops calculation seems a bit more detailed and I am afraid that we are trying to be very detailed compute model for a memory bound op. 
We can simplify and compress the model.

TraceLens/PerfModel/extensions/moe_perf_model_extensions.py
```diff
@@ -1714,3 +1716,253 @@ def get_compute_precision(self):
...
+
+        # Routing fan-out (topk_ids) is the first 2-D int operand [M, topk],
+        # if present (moe_sorting_fwd). Best-effort only.
+        num_tokens = topk = None
+        for shape, dtype in operands:
+            if len(shape) == 2 and dtype == "int":
+                num_tokens, topk = shape[0], shape[1]
+                break
+
+        return {"operands": operands, "num_tokens": num_tokens, "topk": topk}
```
devalshahamd: Looks like num_tokens and topk are not used anywhere, we can reduce the  added code here,

TraceLens/PerfModel/extensions/moe_perf_model_extensions.py
```diff
@@ -1714,3 +1716,253 @@ def get_compute_precision(self):
...
+    def get_maf_type(self):
+        return "vector"
+
+
+class MoeSortScatterGather:
+    """
+    Performance model for the MoE token sort / scatter (permute) auxiliary
+    kernels: aiter::moe_sorting_fwd and aiter::mxfp4_moe_sort_hip.
+
+    These kernels do NOT perform matmuls. They sort the (token, expert)
```
devalshahamd: We can reduce the comments here

devalshahamd (changes requested):
We can reduce some logic and code size here since these are memory bound ops and we do not want to make too many assumptions for counting FLOPs for these operations.

TraceLens/PerfModel/extensions/moe_perf_model_extensions.py
spandoesai: What do you think of simply having 1 FLOP / element? This might also help in other correlation efforts

devalshahamd: That works!

devalshahamd (approved):
LGTM!

ajassani (approved):

PR #768, opened 2026-06-25
TraceDiff Enhancements

kyle-hoffmeyer:
## Summary

This PR enhances TraceDiff's child-alignment algorithm to handle call-stack depth asymmetries between traces and eliminates in-place tree mutation during merging.

---

## Background

TraceDiff merges two call trees by recursively aligning GPU-path children at each node. The upstream algorithm had two steps:

1. **Wagner-Fischer** — edit-distance alignment on GPU-path child lists.
2. **`check_diff_children`** — for unmatched delete/insert pairs, detects when one node is a wrapper around the other (the deleted node's name appears as a child of the inserted node, or vice versa). **Mutates the baseline and variant trees in place** by splicing wrapper nodes out and reparenting their children, then re-runs Wagner-Fischer on the mutated tree.

This worked for simple wrapper differences but had two limitations:

- **Tree mutation** — `check_diff_children` modified `parent["children"]` lists and `child["parent"]` pointers in the original trees. These mutations were permanent side effects visible to all subsequent traversals.
- **No handling of single-child wrapper depth asymmetries** — when one trace wraps operations under an extra stack layer (e.g., `[aten::linear]` containing `[aten::mm, aten::add]` several wrapper layers below vs the other trace exposing `[aten::mm, aten::add]` directly), Wagner-Fischer would classify these as insertions/deletions rather than matches.

---

### 1. New function: `collapse_single_gpu_child`

Descends through nodes that have exactly one GPU-path child, stopping at `cpu_op`, `cuda_runtime`, or nodes with != 1 GPU-path child. Returns `(deepest_uid, gpu_path_children)`. Used by Phases 1 and 4 below.

### 2. `check_diff_children` replaced by `reconcile_unmatched` (no tree mutation)

Same three-case wrapper detection logic:
1. Deleted name appears in inserted node's GPU-path children → substitute inserted UID with its children
2. Inserted name appears in deleted node's GPU-path children → substitute deleted UID with its children
3. Both have identical GPU-path child name lists → substitute both

**Key difference:** Instead of mutating `parent["children"]` and `child["parent"]` in place, `reconcile_unmatched` builds new children lists with substitutions applied and returns them. The original trees are never modified. Children are filtered to GPU-path only and sorted by timestamp.

### 3. Four-phase alignment pipeline in `traverse_and_merge`

The upstream two-step pipeline is replaced with four phases:

**Phase 1: List-level collapse** — if one side has exactly 1 GPU-path child while the other has more, descend the single child through `collapse_single_gpu_child` to find the branch point. Updates `uid1`/`uid2` for the merged event to reflect the actual comparison point. Handles multi-level call-stack depth asymmetries that Phase 3 cannot reach (since `reconcile_unmatched` only checks immediate children):
```
trace1 children: [aten::mm, aten::add]
trace2 children: [aten::linear]
                     └── aten::addmm       (single GPU-path child, descend)
                          └── aten::mm
                          └── aten::add
```
`collapse_single_gpu_child` descends `aten::linear` → `aten::addmm` → stop (2 GPU-path children), producing `[aten::mm, aten::add]` on both sides.

**Phase 2: Alignment** — positional matching (when child counts match and no `cuda_runtime` present) or Wagner-Fischer on original children.

**Phase 3: Reconciliation** — only if unmatched ops remain. Calls `reconcile_unmatched` to detect wrapper nodes among delete/insert pairs where the wrapper has **multiple** GPU-path children. If substitutions are made, re-runs alignment on the substituted children.
```
trace1 children: [aten::mm, aten::relu]
trace2 children: [aten::linear, aten::relu]
                     └── aten::mm
                     └── aten::bias_add
```
W-F matches `relu`/`relu`, leaves `mm` (delete) and `linear` (insert). `reconcile_unmatched` finds "aten::mm" among `linear`'s GPU-path children `[mm, bias_add]`, so it substitutes `linear` with its children. After re-alignment: `mm` matches `mm`, `bias_add` is insert-only. Phase 4's `collapse_single_gpu_child` cannot handle this because `linear` has 2 GPU-path children (it only descends through single-child wrappers).

**Phase 4: Per-node canonicalization** — only if unmatched ops still remain. Collapses each remaining unmatched node individually through **single-GPU-path-child** wrappers via `collapse_single_gpu_child` and re-runs Wagner-Fischer on the collapsed nodes to find additional matches.
```
trace1 children: [..., torch_dispatch, ...]
                       └── aten::mm            (single GPU-path child, descend → stop)
trace2 children: [..., inductor_wrapper, ...]
                       └── compiled_fn         (single GPU-path child, descend)
                            └── aten::mm       (single GPU-path child, descend → stop)
```
W-F sees `torch_dispatch` vs `inductor_wrapper` — names differ, left unmatched. Phase 4 collapses each through its single-child chain: `torch_dispatch` → `aten::mm`, `inductor_wrapper` → `compiled_fn` → `aten::mm`. Both resolve to `aten::mm`, so they are paired.

Tests have shown that all four of these phases are necessary to create the most LCA groupings.

### 4. Graph launch exemption in `subtree_contains_cuda_runtime`

`hipGraphLaunch` and `cudaGraphLaunch` events are now exempt from the `cuda_runtime` check. These are graph replay events, not regular kernel launches, and should not force Wagner-Fischer alignment over positional matching.

### 5. Debug-only `diff_stats` columns

`cpu_op_callstack`, `lca_children_t1`, and `lca_children_t2` are included in `diff_stats` output only when `TRACELENS_DEBUG=1`. These columns are diagnostic aids not consumed by downstream analysis.

---

## Verification

- All 7 trace pairs with same framework, different platforms have same tracediff output before and after. 
- TraceDiff performance has improved on
    - Pair of graph capture augmented vLLM Qwen traces on MI300 and B300
    - Pair of Kimi-K2 traces on different vLLM versions

### DIFF
--- a/TraceLens/PerfModel/extensions/__init__.py
+++ b/TraceLens/PerfModel/extensions/__init__.py
@@ -19,6 +19,7 @@ from .moe_perf_model_extensions import (
     moe_gptq_awq_down,
     moe_flydsl_stage1,
     moe_flydsl_stage2,
+    sglang_fused_append_shared_experts,
 )
 from .attention_perf_model_extensions import (
     InferenceAttention,
@@ -32,6 +33,10 @@ from .perf_model_extensions import (
     gemm_a8w8_blockscale,
     aiter_gelu_and_mul,
     aiter_gelu_tanh_and_mul,
+    aiter_fused_qk_rope_cat_and_cache_mla,
+    sglang_quant_dynamic_mxfp4_quant,
+    aiter_fused_dynamic_mxfp4_quant_moe_sort_hip,
+    aiter_dynamic_per_group_scaled_quant_fp4,
 )
 from .rmsnorm_perf_model_extensions import (
     aiter_rms_norm,
@@ -45,6 +50,7 @@ from .rmsnorm_perf_model_extensions import (
 )
 from .custom_collectives_perf_model_extensions import (
     aiter_fused_allreduce_rmsnorm,
+    aiter_fused_allreduce_rmsnorm_,
     custom_ar_all_reduce,
     aiter_reduce_scatter,
     aiter_all_gather_reg,
@@ -75,6 +81,11 @@ __all__ = [
     "gemm_a8w8_blockscale",
     "aiter_gelu_and_mul",
     "aiter_gelu_tanh_and_mul",
+    "aiter_fused_qk_rope_cat_and_cache_mla",
+    "sglang_quant_dynamic_mxfp4_quant",
+    "aiter_fused_dynamic_mxfp4_quant_moe_sort_hip",
+    "aiter_dynamic_per_group_scaled_quant_fp4",
+    "sglang_fused_append_shared_experts",
     # RMSNorm classes
     "aiter_rms_norm",
     "aiter_rmsnorm",
@@ -86,6 +97,7 @@ __all__ = [
     "vllm_rocm_aiter_triton_add_rmsnorm_pad",
     # Collective classes
     "aiter_fused_allreduce_rmsnorm",
+    "aiter_fused_allreduce_rmsnorm_",
     "custom_ar_all_reduce",
     "aiter_reduce_scatter",
     "aiter_all_gather_reg",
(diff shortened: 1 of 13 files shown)


spandoesai (approved):
LGTM!

PR #769, opened 2026-06-26
Allow csvs and excel to written in one command

kyle-hoffmeyer:
Allows csvs and excel reports to be written in one command (as opposed to only one vs the other). Replicates existing logic in generate_perf_report_pytorch.py


devalshahamd (approved):
LGTM!

PR #770, opened 2026-06-26
Fix EXTRA_VLLM_ARGS and EXTRA_SGLANG_ARGS silently failing

kyle-hoffmeyer:
## Summary

- Added a callout in Step 2b instructing the agent to check which benchmark script Magpie will select before setting `EXTRA_VLLM_ARGS` / `EXTRA_SGLANG_ARGS`
- Added pitfall documenting that InferenceX native scripts silently ignore these variables

## Background

Magpie selects benchmark scripts in priority order (specified here: https://github.com/AMD-AGI/Magpie/blob/main/Magpie/modes/benchmark/benchmarker.py#L1428):
1. User-specified `benchmark_script` in YAML
2. If `benchmark_script` is not specified, searches for InferenceX native script (`gptoss_{precision}_{runner}.sh` for vLLM, `dsr1_{precision}_{runner}.sh` for SGLang)
3. If a script isn't found, uses Magpie generic script (`vllm_{runner}.sh`, `sglang_{runner}.sh`)

Magpie's own generic scripts pass `$EXTRA_VLLM_ARGS` / `$EXTRA_SGLANG_ARGS` through to the server command (ex. https://github.com/AMD-AGI/Magpie/blob/main/Magpie/scripts/benchmark/vllm_mi300x.sh#L92). However,  InferenceX native scripts hardcode their own `vllm serve` invocation and do not reference these variables: \(ex. https://github.com/SemiAnalysisAI/InferenceX/blob/main/benchmarks/single_node/agentic/gptoss_fp4_mi300x.sh#L68C1-L68C20). Thus when a native InferenceX script is selected any flags set in `EXTRA_VLLM_ARGS` are silently ignored with no error or warning.

## Changes

`TraceLens/Agent/Profiling/.cursor/skills/magpie-benchmark-profiling.md`

- **Step 2b, Common vLLM flags:** added a callout block instructing the agent to SSH to the node and check for `gptoss_*` / `dsr1_*` scripts before setting `EXTRA_VLLM_ARGS`. If a native script is found, the agent must stop and ask the user to set `benchmark_script` to a Magpie generic script (e.g. `vllm_mi300x.sh`) before proceeding.
- **Pitfall (new):** documents the silent-ignore behaviour and the fix.


tsrikris:
Looks good but @devalshah-amd is the expert here!

devalshahamd (approved):
LGTM!

tsrikris (approved):

Issue #765, opened 2026-06-25
Root filesystem search when protobuf filename has no directory component crashes Jax reports


brieflynn:
# Issue: `glob.glob` searches root filesystem when protobuf filename has no directory component

In `JaxProfileProcessor.process_protobuf_file`, when `protobuf_file_name` is a bare filename (e.g. `trace.pb` instead of `./trace.pb`), `os.path.dirname` returns an empty string. Concatenating `"/"` produces `"/"`, so `glob.glob` searches from the root filesystem instead of the current working directory.

This causes HLO sidecar `.pb` files to never be found. The JAX report pipeline then fails or produces incomplete output.

**Location:** `TraceLens/util.py`, line 108.

brieflynn:
PR opened with fix https://github.com/AMD-AGI/TraceLens/pull/766

brieflynn:
PR merged

PR #774, opened 2026-06-29
feat(perf): capture output spec + fp8/int8 block-scale config in the unified perf summary

iraj465:
## What
Capture two operand details the unified perf summary was missing — needed for correctness-grade downstream reuse of the per-kernel shapes.

- **fp8/int8 block-scale config**: `gemm_a8w8_blockscale.get_param_details` now emits `quant_scheme`, `quant_granularity`, `block_k`, `block_n`, `scale_dtype`, derived from the captured scale-tensor shapes (`block_k=ceil(K/x_scale_cols)`, `block_n=ceil(N/w_scale_rows)`; verified 128/128). Constant per kernel config ⇒ grouping-neutral; roofline `bytes()` reads only `dtype_A_B` (unchanged).
- **Output spec (inferred)**: the torch/kineto event records input shapes only (`record_shapes`) — there is no output field — so the output is inferred by the perf model (which already computes it for the roofline: GEMM output = M×N). `gemm_a8w8_blockscale` emits `output_shape=(M,N)`, `output_dtype=bf16`. `tree_perf` `args_cols` also carries `Output Dims`/`Output type` for traces that *do* provide them (e.g. JAX), guarded as a no-op otherwise.

`py_compile` clean; please run the TL suite before merge.

### DIFF
--- a/TraceLens/PerfModel/extensions/perf_model_extensions.py
+++ b/TraceLens/PerfModel/extensions/perf_model_extensions.py
@@ -67,18 +67,53 @@ class gemm_a8w8_blockscale(GEMM):
 
     @staticmethod
     def get_param_details(event):
-        return {
+        dims = event["args"]["Input Dims"]
+        types = event["args"].get("Input type", []) or []
+        M, K = dims[0][0], dims[0][1]
+        N = dims[1][0]
+        details = {
             "B": 1,
-            "M": event["args"]["Input Dims"][0][0],
-            "N": event["args"]["Input Dims"][1][0],
-            "K": event["args"]["Input Dims"][0][1],
+            "M": M,
+            "N": N,
+            "K": K,
             "bias": False,
             "dtype_A_B": (
-                event["args"]["Input type"][0],
-                event["args"]["Input type"][1],
+                types[0] if len(types) > 0 else "",
+                types[1] if len(types) > 1 else "",
                 "c10::bfloat16",
             ),
         }
+        # FP8/INT8 block-scale quant CONFIG. Without the scheme + block size a
+        # downstream microbenchmark harness can't build matching scale tensors, so
+        # its reference output mismatches and a real speedup is auto-rejected on
+        # correctness. Derive the block sizes from the captured scale-tensor shapes
+        # (x_scale: (M, ceil(K/block_k)); w_scale: (ceil(N/block_n), ceil(K/block_k)))
+        # and emit the scheme + scale dtype. Constant per kernel config, so this is
+        # grouping-neutral; roofline ``bytes()`` only reads ``dtype_A_B`` (unchanged).
+        try:
+            x_scale, w_scale = dims[2], dims[3]
+            block_k = (-(-K // x_scale[-1])) if x_scale and x_scale[-1] else None  # ceil(K/scale_cols)
+            block_n = (-(-N // w_scale[0])) if w_scale and w_scale[0] else None
+            details.update(
+                {
+                    "quant_scheme": "a8w8_blockscale",
+                    "quant_granularity": "per_block",
+                    "block_k": block_k,
+                    "block_n": block_n,
+                    "scale_dtype": types[2] if len(types) > 2 else "float32",
+                }
+            )
+        except (IndexError, TypeError, ZeroDivisionError):
+            pass
+        # OUTPUT spec is INFERRED, not read from the trace: the torch/kineto event
+        # records input shapes only (record_shapes), so there is no "Output Dims"
+        # field. This kernel's output is the GEMM result Y[M, N] in bf16 (the third
+        # element of dtype_A_B). Emitting it here gives downstream consumers the
+        # output tensor for allocation + correctness without depending on a trace
+        # field that doesn't exist for torch traces.
+        details["output_shape"] = (M, N)
+        details["output_dtype"] = "c10::bfloat16"
+        return details
 
     def bytes(self):
         dtype_A_B = self.param_details["dtype_A_B"]
--- a/TraceLens/PerfModel/extensions/perf_model_extensions.py
+++ b/TraceLens/PerfModel/extensions/perf_model_extensions.py
@@ -86,7 +86,9 @@ class gemm_a8w8_blockscale(GEMM):
         # FP8/INT8 block-scale quant config; block sizes derived from scale-tensor shapes.
         try:
             x_scale, w_scale = dims[2], dims[3]
-            block_k = (-(-K // x_scale[-1])) if x_scale and x_scale[-1] else None  # ceil(K/scale_cols)
+            block_k = (
+                (-(-K // x_scale[-1])) if x_scale and x_scale[-1] else None
+            )  # ceil(K/scale_cols)
             block_n = (-(-N // w_scale[0])) if w_scale and w_scale[0] else None
             details.update(
                 {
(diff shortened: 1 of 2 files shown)


iraj465:
You're right, and thanks for catching it — verified: the torch/kineto event records **input shapes only** (`record_shapes`), so there's no output field. `"Output Dims"` is read in exactly one place — `perf_model.py:4189`, the **JAX conv** path — and the real `unified_perf_summary.csv` has no `Output Dims` column. So adding it to `args_cols` is a **no-op for torch traces** (the `if arg in columns` guard skips it). My mistake in framing it as "the trace records them."

Correct approach (pushed): **infer** the output from the perf model, which already computes it for the roofline (`bytes_output = M*N*bpe_output`). `gemm_a8w8_blockscale.get_param_details` now emits `output_shape=(M,N)`, `output_dtype=bf16`. I left the `args_cols` carry in place (guarded) so JAX traces that *do* provide output dims still benefit, but the torch path no longer depends on a non-existent field.

The general rollout — per-op output inference for the non-GEMM models (attention, elementwise, etc.) — is your domain since each perf model knows its own output (e.g. CONV already has `get_output_shape`). Happy to do the GEMM/MoE ones first and follow whatever pattern you prefer for the rest.

The fp8 block-scale config part (`block_k/block_n/scheme/scale_dtype`) is unaffected — it's derived from the captured scale-tensor `Input Dims`, which the torch trace does carry.

TraceLens/TreePerf/tree_perf.py
```diff
@@ -720,7 +720,17 @@ def summarize_df_perf_metrics(
             )
             if "overlap_pct" in df_perf_metrics.columns:
                 dict_agg["overlap_pct"] = agg_metrics
-        args_cols = ["Input Dims", "Input type", "Input Strides", "Concrete Inputs"]
+        # Carry OUTPUT dims/type when a trace actually provides them (e.g. JAX
+        # traces set ``event["args"]["Output Dims"]``). NOTE: torch/kineto traces
+        # record INPUT shapes only (``record_shapes``) — there is no output field —
+        # so for those workloads the output spec is INFERRED by the perf model
+        # instead (see e.g. gemm_a8w8_blockscale.get_param_details ->
+        # output_shape/output_dtype). The ``if arg in columns`` guard makes this a
```
devalshahamd: How are Output Dims and type popluated here?

TraceLens/TreePerf/tree_perf.py
```diff
@@ -720,7 +720,17 @@ def summarize_df_perf_metrics(
...
+        # safe no-op when the trace lacks the columns.
+        args_cols = [
```
devalshahamd: Let's not add this to the main tree_perf.py yet. We can add it as param_details in for this particular op.

iraj465:
Thanks Deval — this is all very helpful. Responses:

**1) #774 / output dims.** Agreed on both points:
- I **reverted the `tree_perf.py` change** (pushed) — for torch traces it was a no-op anyway, and not worth the cross-workload risk. Output now comes purely from the perf model's `get_param_details`, which already computes the output for the roofline (`bytes_output = M*N*bpe_output`), so we just *surface* `output_shape`/`output_dtype` there.
- On "the key feature that covers all ops, not just one": exactly right. The clean general rule is **each perf model emits its inferred output (shape+dtype)** — they all already know it for the bytes calc (GEMM→(M,N), CONV has `get_output_shape`, elementwise→broadcast). So the input side needs **nothing new** (Unified_perf_summary already has Input Dims/types/Concrete Inputs); the only missing piece for a correct reference is the **inferred output**. Happy to do GEMM/MoE in `perf_model_extensions.py` and have you extend the pattern to the base classes you own.
- To be concrete on "what info is required for a correct reference": (a) input shapes+dtypes [have], (b) inferred output shape+dtype [this PR], (c) for value-dependent ops, the value distribution [your event-replay, see #3]. That's the full set.

**2) Per-regime / weighting.** You're right that analysis.md carries shapes+counts. The concrete gap I see: `call_count` in `kernel_candidates.json` is populated for *some* kernels but **0 for the multi-shape composite candidates** — e.g. the GLM-4.7 fused-MoE candidates (k012/13/14) have `call_count=0` with 21 shapes, while k010/11 have `call_count=2852`. Those 0-count composites are exactly the ones that need per-shape weighting. So +1 to "fix at the candidate + dispatch layer to include weights based on call_count" — that's the real fix.
- On the prefill/decode **label**: for GEAK the **weights (per-shape call_count) are what matter** — it time-weights the shape set, no label needed. The prefill/decode label is only nice-to-have for reporting / human reasoning ("this opt helps decode"). So please prioritize the per-shape count/weight; the label is optional.

**3) Value distributions (#765).** `EventReplay/custom_inits.py` on `feature/event-replay-custom-ops` is exactly what we need — `MoeRoutingInit` (uniform/zipf for `ck_moe_stage1/2`) and `PagedAttentionInit`. We'll have the harness draw from those instead of uniform-random for the value-dependent ops. One note: that covers the aiter CK MoE; the sglang **triton** fused-MoE will want an analogous routing init — can reuse the same distribution logic. Let's sync today on fp8 scale ranges.

**4) kernel_details_summary.** Great, thanks for confirming — that per-device-kernel duration column is what lets us pin composite ops (fused-MoE, gdn-attention) to the single hot source. Glad it stays.

Net: this makes any optimizer (OOB / GEAKv3 / GEAKv4) get the full input+output+quant picture from the real trace → more E2E-faithful. I'll keep #774 scoped to perf-model output inference + fp8 config; the regime-weighting + value-distribution wiring lands on our side (candidate/dispatch + harness).

TraceLens/PerfModel/extensions/perf_model_extensions.py
```diff
@@ -67,18 +67,53 @@ class gemm_a8w8_blockscale(GEMM):
...
+                }
+            )
+        except (IndexError, TypeError, ZeroDivisionError):
+            pass
+        # OUTPUT spec is INFERRED, not read from the trace: the torch/kineto event
+        # records input shapes only (record_shapes), so there is no "Output Dims"
+        # field. This kernel's output is the GEMM result Y[M, N] in bf16 (the third
+        # element of dtype_A_B). Emitting it here gives downstream consumers the
+        # output tensor for allocation + correctness without depending on a trace
+        # field that doesn't exist for torch traces.
```
devalshahamd: Compress to 1 line

TraceLens/PerfModel/extensions/perf_model_extensions.py
```diff
@@ -67,18 +67,53 @@ class gemm_a8w8_blockscale(GEMM):
...
                 "c10::bfloat16",
             ),
         }
+        # FP8/INT8 block-scale quant CONFIG. Without the scheme + block size a
+        # downstream microbenchmark harness can't build matching scale tensors, so
+        # its reference output mismatches and a real speedup is auto-rejected on
+        # correctness. Derive the block sizes from the captured scale-tensor shapes
+        # (x_scale: (M, ceil(K/block_k)); w_scale: (ceil(N/block_n), ceil(K/block_k)))
+        # and emit the scheme + scale dtype. Constant per kernel config, so this is
+        # grouping-neutral; roofline ``bytes()`` only reads ``dtype_A_B`` (unchanged).
```
devalshahamd: Convert it to 1-line comment

devalshahamd (approved):
LGTM!

devalshahamd (approved):

PR #777, opened 2026-06-30
Set idle threshold as percent

kyle-hoffmeyer:
Previously if the GPU had VRAM usage of over 256 MB, it was flagged as not being idle. This caused issues on large systems with lots of memory where 256 MB of VRAM is very little.
Now, if the GPU's VRAM is over 10% occupied, it is flagged as not being idle.

The new python API calls to rocm-smi and nvidia-smi have been verified and tested on MI300 and B300.


mohbasit (approved):
looks good

gabeweisz (approved):

PR #780, opened 2026-06-30
perf: add byte model for aiter::_gemm_a16w16_asm (Switch-Gate)

spandoesai:
The ASM variant of gemm_a16w16_atomic_ has a different argument layout: the output tensor is at Input index 2, while index 3 is a split-K workspace recorded as 'unsigned int'. The shared gemm_a16w16_atomic_ model reads the output dtype from index 3, so name2bpe returns None and GEMM.bytes_func bails out, surfacing as Data Moved = NaN for this kernel (e.g. the DeepSeek-R1 MoE Switch-Gate GEMM in decode).

Add a dedicated gemm_a16w16_asm subclass that reads the output dtype from the real output tensor (index 2, falling back to the activation dtype), and re-point the aiter::_gemm_a16w16_asm mapping to it. The kernel now reports Data Moved = (M*K + K*N + M*N)*bpe instead of NaN.


TraceLens/PerfModel/extensions/perf_model_extensions.py
```diff
@@ -433,6 +433,45 @@ def get_param_details(event):
...
+
+    Why this needs its own get_param_details:
+        The ASM kernel's argument layout differs from the Triton atomic
+        variant.  The output tensor is at Input index 2, while index 3 is a
+        split-K workspace recorded with dtype 'unsigned int' (not a tensor
+        dtype).  The parent reads the output dtype from index 3, so
+        name2bpe('unsigned int') returns None and GEMM.bytes_func bails out
+        (returning None), which surfaces as Data Moved = NaN for this kernel.
+        Read the output dtype from the real output tensor (index 2) instead,
+        falling back to the activation dtype when it is missing/unknown.
```
devalshahamd: We can shorten this string a bit.

devalshahamd (commented):
Just one minor comment, otherwise LGTM!

devalshahamd (approved):
LGTM!

ajassani (approved):

PR #786, opened 2026-07-02
Generic trace splitting support

kyle-hoffmeyer:
## Motivation

This PR adds support for splitting non-vLLM nor SGLang traces. `split_inference_trace_annotation.py` previously only supported traces from vLLM and SGLang. These frameworks annotate each execution step with a structured name (e.g. `execute_4_context_3(sq128sk256...)_generation_2(...)`) that the tool parses to identify iteration boundaries, extract per-step traces, and detect steady-state regions.

Traces from other workloads carry none of these annotations and trace splitting fails.

The goal of this change is to make the splitting and steady-state extraction pipeline work for any workload that has a repeating loop structure in its call tree, without requiring framework-specific annotation support.

## Summary

For vLLM/SGLang traces, iteration boundaries are identified by matching structured annotation names (e.g. `execute_4_context_3(...)_generation_2(...)`) injected by the framework at runtime. Each annotation event spans exactly one execution step; its `ts` and `ts + dur` define the window.

## Implementation

The tool splits a large trace into per-iteration traces by finding events that mark the start and end of each iteration, then extracting all CPU and GPU events that fall within each iteration's time window. For generic traces (this PR), iteration boundaries are identified by building a call tree with Trace2Tree and finding the shallowest node whose children form a repeating sequence. Each repeating block becomes a synthetic root event spanning from the first to the last child of that block. Once iteration roots are found — regardless of how they were detected — the same downstream pipeline of identifying the steady state windows and extracting the iterations occurs.

### Entry point: `find_iteration_roots`

The existing function tries annotation patterns in priority order. This PR adds a third tier that fires only when both the primary (vLLM 1211/1213 pattern) and backup (SGLang `step[...]` pattern) return nothing:

```
primary annotation patterns
    → backup annotation patterns
        → generic call-tree fallback (new)
```

### Finding the repeating pattern: `_detect_iteration_roots_from_tree`

The goal is to find the highest event in the call stack that has a list of children that appear in a repeating pattern. From the root, a BFS search occurs. At each event, the list of children is checked to see if it contains a meaningful pattern. The first such list of children forms the iterations.

Nodes with no GPU-bearing children are skipped entirely.

**Period detection: `_find_repeating_period`.** The full child name list is passed to this function, which finds the shortest name sequence that repeats consecutively: the iteration pattern.

It works by sliding a `start` offset forward from 0. At each `start`, it takes the sub-list `names[start:]` and tries candidate period lengths `p = 1, 2, 3, ...`. For each `p`, it treats `names[start : start + p]` as the candidate pattern and counts how many times that exact block appears back-to-back starting at `start`. A period is accepted when:

- It repeats at least 3 consecutive times (guards against coincidental repetitions in short utility call lists), and
- The repeating portion covers more than 50% of `names[start:]` (guards against a short pattern that only matches a small prefix).

The `start` offset exists to skip setup events that appear before the loop body.

For trace at `tests/traces/h100/Wan-AI_Wan2.1-T2V-1.3B-Diffusers__1016009.json.gz`, the GPU-path children of `__call__` (after skipping the setup prefix) look like:

```
to, expand, WanTransformer, do_classifier_free_guidance, WanTransformer, sub, mul, add, scheduler.step, len, tqdm::update, interrupt, __setattr__,   ← iter 0
to, expand, WanTransformer, do_classifier_free_guidance, WanTransformer, sub, mul, add, scheduler.step, len, tqdm::update, interrupt, __setattr__,   ← iter 1
...
```

The function finds `p=13` with that full 13-event sequence repeating, and returns `(13, pattern, start_offset)`.

**Iteration grouping.** Once the period and pattern are found, each iteration is bounded by the Nth occurrence of the first event in the pattern (`to`) and the Nth occurrence of the last event in the pattern (`__setattr__`).

**Synthetic root events.** Each block is converted to a synthetic root event by copying the first event in the block and overwriting its `dur` to span from `block[0].ts` to `block[-1].ts + block[-1].dur`. This is the format `extract_iteration` expects: a single event whose time window covers the full iteration.

## Testing

Validated against:

- `tests/traces/h100/Wan-AI_Wan2.1-T2V-1.3B-Diffusers__1016009.json.gz`:
Trace:
image

After splitting:
image

- `tests/traces/mi300/google_owlv2-large-patch14-ensemble__1016001.json.gz`:
Trace:
image

After splitting:
image

- All 7 existing unit tests in `tests/test_split_inference_trace_annotation.py` continue to pass. The new fallback is only triggered when both annotation pattern tiers return nothing, so existing vLLM/SGLang behaviour is unchanged.

### DIFF
--- a/TraceLens/TraceUtils/split_inference_trace_annotation.py
+++ b/TraceLens/TraceUtils/split_inference_trace_annotation.py
@@ -302,13 +302,13 @@ def _detect_iteration_roots_from_tree(
     ``roots`` may be a single event dict or a list of event dicts — all are
     seeded into the BFS at depth 0 so they are explored level-by-level together.
 
-    Pattern detection uses only GPU-path children (dur > 1us) to avoid noise
-    from utility calls. Block boundaries are expanded to the full unfiltered
-    child list to include CPU-only events (e.g. scheduler.step) that follow
-    the GPU work within each iteration.
+    Pattern detection uses all children (not just GPU-path ones) so that
+    leading CPU-only events (e.g. ``next`` in the OWL pipeline) are included
+    as part of the iteration anchor. A minimum child count guards against false
+    positives from short utility-function child lists.
 
     Returns a list of synthetic root events, one per detected iteration, where
-    each event's ``dur`` spans the full block from first to last child.
+    each event's ``dur`` spans from the first to the last child of the block.
     """
     from collections import deque
 
@@ -322,18 +322,12 @@ def _detect_iteration_roots_from_tree(
         if not children:
             continue
 
-        # Use GPU-path children (dur > 1us) for pattern detection to avoid noise
-        # from utility calls. Skip nodes with no qualifying children — recurse
-        # only into GPU-bearing subtrees.
-        gpu_children = [
-            c for c in children
-            if c.get("gpu_events") and c.get("dur", 0) > 1
-        ]
-        if not gpu_children:
+        # Only recurse into GPU-bearing subtrees.
+        if not any(c.get("gpu_events") for c in children):
             continue
 
         p, _, start = _find_repeating_period(
-            [c.get("name", "") for c in gpu_children]
+            [c.get("name", "") for c in children]
         )
         if p is None:
             for child in children:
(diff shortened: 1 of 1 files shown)


devalshahamd (approved):
LGTM!

PR #787, opened 2026-07-02
fix(jax): parse integer HLO lines and resolve collective aliases

Ahmedhasssan-aig:
## Summary
- Addressed the issue raised "https://github.com/AMD-AGI/TraceLens-internal/issues/527"
- Extend HLO graph parsing to include integer and boolean element types (`s32`, `s64`, `u32`, `u64`, `pred`) so trace metadata is not dropped for non-float ops
- Map runtime collective names (e.g. `reduce-scatter.12`) to HLO dump keys (`reduce-scatter-start`) when linking GPU events to HLO metadata
- Deduplicate missing-`hlo_op` warnings (log once per module+op)
- Preserve FA FWD / FA BWD / FA V3 kernel categories when `hlo_op` would incorrectly classify them as TE

## Motivation
On JAX profiles (e.g. `tw052.xplane.pb`), the legacy HLO parser only matched float element types, causing thousands of `Missing hlo_op` warnings and missing `op_gemm` / `op_te` xlsx sheets. Async collectives also use different names in traces vs HLO dumps.

## Files changed
- `TraceLens/util.py`
- `TraceLens/Trace2Tree/trace_to_tree.py`


devalshahamd (approved):
LGTM!

ajassani (approved):

PR #789, opened 2026-07-06
Propagate standalone table cols to comparative

kyle-hoffmeyer:
Adds the Kernel Path and Kernel Name columns to comparative tables. Fixes oversight in `validation_utils.py` that wasn't checking for `Args (T1)` column in comparative mode. It was previously only checking for `Args` column when in standalone mode.


tsrikris (approved):

tsrikris:
LGTM!

PR #795, opened 2026-07-08
Add kernel name based matching for multistream run

devalshahamd:
This pull request introduces improved support for merging HIP graph traces involving multiple streams in the `trace_capture_merge_experimental.py` script. The main changes add logic to align capture dispatches to graph kernel executions when kernels are distributed across multiple streams, enhancing the robustness and accuracy of the trace merging process.

**Multistream graph alignment:**

* Added helper functions (`_stream_of`, `_names_match`, `is_multistream`, `capture_has_kernel_names`, `_best_stream_by_run`, and `align_streams`) to support matching and aligning capture events to graph events across multiple CUDA streams.
* Updated the merge logic to detect multistream graph roots and, if present, use kernel names and stream information to align events; if alignment fails or kernel names are missing, a warning is issued and the merge is marked as failed.
* Introduced a `merge_failed` flag to track failures during the merge process, and return `None` if any failure occurs, ensuring the report is not augmented with incorrect data. [[1]](diffhunk://#diff-a361820c09001fd17b2d09215e178075fa4e355e907c434103b7f0bd0a40041cR637) [[2]](diffhunk://#diff-a361820c09001fd17b2d09215e178075fa4e355e907c434103b7f0bd0a40041cR752-R756)

**Dependency update:**

* Imported `deque` from `collections` to support per-stream FIFO queueing in the new alignment logic.

### DIFF
--- a/TraceLens/Trace2Tree/trace_capture_merge_experimental.py
+++ b/TraceLens/Trace2Tree/trace_capture_merge_experimental.py
@@ -115,7 +115,12 @@ def _names_match(c_event, g_name):
 
 def is_multistream(graph_filtered_events):
     """True when the graph root's kernels span more than one CUDA stream."""
-    return len({_stream_of(e) for e in graph_filtered_events}) > 1
+    streams = set()
+    for e in graph_filtered_events:
+        streams.add(_stream_of(e))
+        if len(streams) > 1:
+            return True
+    return False
 
 
 def capture_has_kernel_names(capture_filtered_events):
--- a/TraceLens/Trace2Tree/trace_capture_merge_experimental.py
+++ b/TraceLens/Trace2Tree/trace_capture_merge_experimental.py
@@ -17,7 +17,7 @@ model suitable for performance analysis.
 
 import sys
 import time
-from collections import OrderedDict
+from collections import OrderedDict, deque
 from typing import Any, Dict, List, Optional, Tuple
 import json
 import os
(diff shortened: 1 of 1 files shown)


TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -99,6 +99,85 @@ def _align_capture_to_graph(capture_events, graph_events):
...
 
 
+def _stream_of(event):
+    """Return the CUDA stream id a graph kernel executed on, or None."""
+    return event.get("args", {}).get("stream")
+
+
+def _names_match(c_event, g_name):
+    """Pair a capture dispatch to a graph kernel by name (Memcpy/Memset carry no kernel arg)."""
+    if ("Memcpy" in c_event["name"] and "Memcpy" in g_name) or (
```
kyle-hoffmeyer, line 109: We could have a helper function that checks if an event is Memcpy or Memset since this is checked in many places now.

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -617,17 +697,35 @@ def merge_capture_trace_into_graph(
...
-                print(
-                    "Warning: subtree events verification failed for capture root {} and graph root {}".format(
-                        c_root["name"], g_root["name"]
+            if is_multistream(graph_filtered_events):
+                if not capture_has_kernel_names(capture_filtered_events):
+                    print(
+                        "Graph capture merge failed: multistream graph root {} but "
+                        "capture kernel names are absent".format(g_root["name"])
+                    )
+                    merge_failed = True
```
kyle-hoffmeyer, line 711: With this condition, the merge fails if a single capture event doesn't have kernel information. Should we not instead skip the capture events without kernel info? It could be that rocprofiler failed to capture kernel info for just a couple of events rather than all of the events.

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -99,6 +99,85 @@ def _align_capture_to_graph(capture_events, graph_events):
...
+    if ("Memcpy" in c_event["name"] and "Memcpy" in g_name) or (
+        "Memset" in c_event["name"] and "Memset" in g_name
+    ):
+        return True
+    return _capture_kernel_name(c_event) == g_name
+
+
+def is_multistream(graph_filtered_events):
+    """True when the graph root's kernels span more than one CUDA stream."""
+    return len({_stream_of(e) for e in graph_filtered_events}) > 1
```
kyle-hoffmeyer: This could be made more efficient to just return true once it encounters two events with different streams. Not necessary though

kyle-hoffmeyer (commented):
Two small comments and one concern.

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
devalshahamd: added an early exit condition

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
devalshahamd: It is tricky in multistream trace to merge without kernel names, so this is a bit conservative but robust approach.  We can relax this a bit once we have stream information in the capture trace so we don't rely on just names.

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
devalshahamd: Good point. We can create an issue for now to refactor the code...

kyle-hoffmeyer (approved):

PR #798, opened 2026-07-09
Robustness to misordering between graph and capture events

kyle-hoffmeyer:
## Summary

Graph replay events sometimes appear in different order from capture events. The reasons for this are not clear and is something that needs to be investigated.

For one trace the events were ordered as follows:

```
Capture cycle:         NORM  G128  ROPE  KVC  ATTN_QKV  ATTN_RED  OTHER  NORM  G512  ACT  G256
Graph replay cycle: ACT   G256  NORM  G128  ROPE  KVC  ATTN_QKV  ATTN_RED  OTHER  NORM  G512
```

The activation in some attention->FFN cycles appeared at the start in some of the graph replay cycles which is incorrect. However, the count of each kernel invocation is the same between the two traces.

To account for this, there is now a phase 3 fallback. After normal fully positional graph capture merge and greedy graph capture merge fails, this new attempt at merging is used. 

## Approach

Group both event lists by kernel name, then match positionally within each group. Reorder **graph replay** events to match **capture** order (not the reverse), so the merged tree preserves the logical graph-node execution structure.

### Algorithm

1. **Group by name**: Build `{kernel_name: [event, ...]}` for both capture and graph events, preserving within-group order.

2. **Verify counts**: For every kernel name, check that the count in capture equals the count in graph. If any name has a mismatch, return `None` (alignment failed).

3. **Reassemble in capture order**: Walk the capture event list. For each capture event, pull the next unused graph event with the same kernel name from its group. The result is the graph events reordered to match capture (logical) order.

```
capture (CPU dispatch order):  [A0, B0, C0, A1, B1, A2]
graph   (GPU execution order): [A0, C0, B0, A1, B1, A2]

Step 1 - group by name:
  capture_groups: { A: [A0, A1, A2],  B: [B0, B1],  C: [C0] }
  graph_groups:   { A: [A0, A1, A2],  B: [B0, B1],  C: [C0] }

Step 2 - counts match: A=3, B=2, C=1  OK

Step 3 - walk capture order, pull from graph groups:
  capture[0] = A  -> graph_groups[A][0] = A0
  capture[1] = B  -> graph_groups[B][0] = B0
  capture[2] = C  -> graph_groups[C][0] = C0
  capture[3] = A  -> graph_groups[A][1] = A1
  capture[4] = B  -> graph_groups[B][1] = B1
  capture[5] = A  -> graph_groups[A][2] = A2

Result: graph events reordered to [A0, B0, C0, A1, B1, A2]
        -> matches capture order
        -> capture[i] <-> graph[i] pairing is correct
```


TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -99,6 +99,45 @@ def _align_capture_to_graph(capture_events, graph_events):
     return aligned
 
 
+def _align_graph_to_capture_by_group(capture_events, graph_events):
```
devalshahamd, line 102: So this will resolve the issue if the kernel appears out of order right?

devalshahamd (approved):
LGTM!

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
kyle-hoffmeyer: Correct

PR #811, opened 2026-07-14
FOR HL V0.9: Remove PID check in microbenchmark

kyle-hoffmeyer:
Remove PID check when determining whether GPU is idle. Utilization and memory checks are enough.


gabeweisz:
Why don't we want to check PIDs? Is there an argument for making this a flag that defaults off rather than removing the code?

kyle-hoffmeyer:
Commit has been cherry-picked

PR #820, opened 2026-07-17
TraceDiff: Aggressive WF name matching during phase 4

kyle-hoffmeyer:
## Background

In phase 4 of matching, TraceDiff takes the children that have already been passed through wagner fischer and filters them to only the unmatched children. It then collapses the children down to the first event with more than one child and performs wagner fischer again on just these unmatched children.

## Summary
This PR adds a final wagner fischer at the end of phase 5 to more aggressively match events. If two events are python functions but slightly differ in name, `_normalize_name_for_comparison` now strips the function name as well as the path leading to the file.

## Details
If the event name is
```
vllm/model_executor/layers/fused_moe/gpt_oss_triton_kernels_moe.py(148): legacy_routing_from_sparsematrix
```

_normalize_name_for_comparison(event_name, strip_details=True) now changes event name to be:
```
gpt_oss_triton_kernels_moe.py
```

The regex pattern only applies to python functions. This more aggressive version of wagner fischer is only applied at the end of phase 4, to avoid matching unrelated events in the first wagner fischer passes when there may be better matches. 

For instance:
```
trace1 (collapsed children):        trace2 (collapsed children):
  [0] kernels.py: compute_attention   [0] kernels.py: compute_ffn
  [1] kernels.py: compute_ffn         [1] kernels.py: compute_attention
```
If we do aggressive WF, children become:
```
trace1: [kernels.py, kernels.py]
trace2: [kernels.py, kernels.py]
```
and kernels are matched positionally:
```
[0] kernels.py  ↔  [0] kernels.py   →  compute_attention ↔ compute_ffn     ✗ WRONG
[1] kernels.py  ↔  [1] kernels.py   →  compute_ffn ↔ compute_attention     ✗ WRONG
```

Therefore an initial normal WF pass occurs first, followed by a the more aggressive WF pass on the remaining unmatched events.


gabeweisz (approved):

spandoesai (approved):
LGTM

PR #825, opened 2026-07-17
Docs/updates

mattwill-amd:
Manager feedback

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -22,48 +22,50 @@ automated GPU trace analysis, available as command-line tools and a Python SDK.
 
 ### Release highlights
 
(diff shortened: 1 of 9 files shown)


ajassani:
One note on `compatibility.md`: this reorder moves **Hardware requirements (roofline analysis)** above **ROCm and CUDA compatibility**. I'd prefer ROCm/CUDA stay first and the hardware section not lead the page.

Context: TraceLens runs on CPU — it parses trace files and never touches the GPU — so its actual hardware requirement is just a CPU host. The "Hardware requirements (roofline analysis)" table really describes the *tracing environment* / target accelerator being profiled, not what's needed to run TraceLens. Leading with it reads as "you need an MI300X to use TraceLens," which isn't the case.

No need to block this PR on it — I'll clear this up (restore the ordering and reword the section to make the distinction clear) in a follow-up PR.

ajassani (approved):

PR #828, opened 2026-07-20
Local Evals

gabeweisz:
- Modify the repeatability test script to be able to run against a local Pi server
- Create a script to run a local server (usually vllm but sglang should work), install and configure Pi, and run local evals using it


Ahmedhasssan-aig (approved):
LGTM!

ajassani (approved):

PR #835, opened 2026-07-21
TraceDiff: Fix name disambiguation and add tests

kyle-hoffmeyer:
## Summary

This PR rewrites `_disambiguate_same_name_candidates` in to fix three correctness issues with how same-named events are matched between traces. Fixes issues in https://github.com/AMD-AGI/TraceLens/pull/800.

#### Issue 1: Matched events excluded from candidate pool

**Problem:** The old implementation only considered unmatched events (inserts/deletes) as candidates for reassignment. Already-matched events with the same name were never considered. This meant that if WF produced two incorrect matches, the function could never correct them.

**Fix:** Candidate pools now include ALL same-named events on each side -- both matched and unmatched.

#### Issue 2: Same candidate assigned to multiple matches

**Problem:** The old implementation iterated over each match op independently, evaluating candidates for that single match in isolation. If two different match ops both identified the same unmatched candidate as the correct match, that candidate could be reassigned to both -- producing duplicate trace2 (or trace1) indices in the final match list.

**Fix:** The new implementation operates per-name-group, scoring all `(i, j)` pairs at once and greedily assigning them from a single sorted list. Each index can only be assigned once.

#### Issue 3: Wrong side left unmatched

**Problem:** Consider 5 events in trace1 and 3 events in trace2, all with the same name. WF aligns positionally: trace1[0, 1, 2] get matched, trace1[3, 4] become deletes. The old code iterated per-match-op and tried to swap in unmatched candidates, but there was no guarantee that the *correct* trace1 events were left unmatched. Since trace2 has fewer events (3 < 5), all 3 trace2 events should be matched with *some* trace1 event, but the old logic couldn't enforce this.

**Fix:** The function now operates per-name-group rather than per-match-op. For each ambiguous name, it collects all indices on both sides and produces exactly `min(n1, n2)` matches using a greedy algorithm scored by BFS subtree match depth. This guarantees the minority side (the side with fewer events of that name) is fully matched. Only majority-side leftovers become deletes/inserts.

## Algorithm

For each normalized name that appears on both sides with ambiguity (>1 on at least one side) and participates in at least one existing match:

1. Collect all trace1 and trace2 indices with that name (matched + unmatched)
2. Precompute BFS-level subtree fingerprints for each node
3. Score every `(i, j)` pair as `(full_levels, partial_overlap)` where `full_levels` is the number of BFS levels with identical child-name tuples before the first divergence, and `partial_overlap` is the multiset intersection size at that first diverging level
4. Sort pairs by score descending, breaking ties by: prefer existing WF match, then positional order
5. Greedily assign `min(n1, n2)` non-conflicting pairs from the sorted list
6. Remaining majority-side nodes become deletes/inserts

## Tests
- `test_no_ambiguity_returns_ops_unchanged` -- No same-named duplicates; verifies ops are returned as-is
- `test_unique_survivor_trace2_triggers_reassignment` -- WF matched trace1[0] to trace2[0], but trace2[1] has an identical subtree; verifies reassignment to the better candidate
- `test_multiple_survivors_keeps_original_wf_match` -- Two trace2 candidates with identical (empty) subtrees; verifies WF's original choice is preserved when the tie cannot be broken
- `test_no_duplicate_trace2_index_in_matches` -- 3 trace1 vs 4 trace2 events, all same name; verifies no trace2 index appears in more than one match
- `test_no_duplicate_trace1_index_in_matches` -- 4 trace1 vs 3 trace2 events, all same name; verifies no trace1 index appears in more than one match
- `test_insert_and_delete_preserved_when_no_reassignment` -- Unambiguous inserts and deletes pass through unchanged
- `test_multiple_same_name_different_subtrees_correct_matches` -- 2 trace1 vs 3 trace2, each with a distinct kernel child; verifies a wrong positional match is corrected to the right subtree pairing
- `test_second_match_falls_back_when_best_candidate_already_claimed` -- Two trace1 events both prefer the same trace2 candidate; verifies the second gets its next-best available match
- `test_swapped_matches_corrected_via_matched_candidates` -- WF produced two all-matched events with names swapped; verifies matched events are now candidates (Issue 1)
- `test_minority_side_fully_matched` -- 5 trace1 vs 3 trace2 events, all same name; verifies all minority-side events are matched and the correct trace1 events are left unmatched (Issue 3)
- `test_non_positional_match_across_sides` -- 5 vs 3 events where two trace1 events share BFS level-1 with a trace2 event but only one matches at level-2; verifies the deeper match wins
- `test_partial_overlap_breaks_tie_at_diverging_level` -- Two trace1 events diverge from the same trace2 event at the same BFS depth but with different overlap counts; verifies partial overlap correctly breaks the tie
- `test_all_ops_indices_valid` -- All match/delete/insert indices in the result are within bounds


spandoesai (approved):
LGTM! This PR adds tests for name disambiguation, along with some other fixes

PR #848, opened 2026-07-24
Add TraceLens xDiT profiling support

kyle-hoffmeyer:
## Summary

Add TraceLens profiling support for xDiT diffusion using similar capture-merge infrastructure as vLLM/SGLang. A patch applied to xDiT's base model captures per-kernel shapes and call stack information during the graph capture phase, which TraceLens merges into the graph-replay timing trace for roofline analysis.

### Patch (`config_xdit_v26.6.patch`, `config_xdit_v26.7.patch`)

Two versioned patches targeting `rocm/pytorch-xdit:v26.6` and `rocm/pytorch-xdit:v26.7` respectively.

**`xfuser/config/args.py`**: Adds `profile_capture_phase: bool = False` field to `xFuserArgs` (xDiT's main config class) and `--profile_capture_phase` CLI flag.

**`xfuser/model_executor/models/runner_models/base_model.py`**: Added `_run_compile_warmup()` method to handle graph capture profiling logic.

Call stack before: `_compile_model() -> _run_timed_pipe()` 
Call stack with patch: `_compile_model() -> _run_compile_warmup()  -> _run_timed_pipe()`  

When `--profile_capture_phase` is enabled and conditions are met, this method splits the compile warmup into two phases:

```python
def _run_compile_warmup(self, compile_args: dict) -> None:
    enable_capture = (
        self.config.profile_capture_phase
        and get_world_group().rank == 0
        and self._get_compile_mode() == "reduce-overhead"
        and not self.pipe.scheduler.config.shift_terminal
    )
    if not enable_capture:
        self._run_timed_pipe(compile_args)
        return

    # Phase 1: compile + eager warmup (1 step, no profiler)
    compile_args["num_inference_steps"] = 1
    self._run_timed_pipe(compile_args)

    # Phase 2: CUDA graph recording (1 step, with profiler)
    with profile(record_shapes=True, with_stack=True) as capture_prof:
        self._run_timed_pipe(compile_args)
    capture_prof.export_chrome_trace(...)
```

#### `enable_capture` conditions
- **`self.config.profile_capture_phase`**: The `--profile_capture_phase` CLI flag must be explicitly set.
- **`get_world_group().rank == 0`**: Only rank 0 collects the capture trace, matching vLLM/SGLang behavior.
- **`self._get_compile_mode() == "reduce-overhead"`**: Capture profiling is only needed for models compiled with CUDA graphs (`reduce-overhead` mode). In `default` mode, there are no CUDA graphs — the profiler already sees individual kernel launches during the replay trace, so no separate capture phase is needed.

- **`not self.pipe.scheduler.config.shift_terminal`**: Models whose diffusers scheduler sets `shift_terminal` cannot run with `num_inference_steps=1` which is required for graph capture profiling. Bug tracked https://github.com/huggingface/diffusers/issues/14411.
#### Why the warmup is split into two phases

In xDiT, `torch.compile(mode="reduce-overhead")` uses a three-phase warmup:
1. **Phase 1** (eager warmup): Dynamo traces the forward pass, Inductor compiles and generates kernels, runs eagerly.
2. **Phase 2** (CUDA graph recording): The compiled kernels are recorded into a CUDA graph, capturing individual kernel launches.
3. **Phase 3+** (CUDA graph replay): A single `cudaGraphLaunch` replays all kernels — individual launches are invisible to the profiler.

**`xfuser/model_executor/models/runner_models/ltx.py`**: Both `xFuserLTX23VideoModel` and `xFuserLTX2VideoModel` have their own `_compile_model()` overrides. These are updated to call `_run_compile_warmup()` instead of `_run_timed_pipe()`.

**`xfuser/core/distributed/attention_backend.py`** (v26.6 only): The v26.6 patch checks out a newer xDiT commit than what's in the image, requiring AITER sage attention custom_op wrappers to be reapplied. This is not needed in v26.7 as the patch applies directly to the image's xDiT without checkout.

### Docker Build Script (`build_docker_xdit.sh`)

Usage:
```bash
bash build_docker_xdit.sh v26.7 /path/to/TraceLens -t tracelens-xdit:v26.7
```

Supports both v26.6 and v26.7 versions. Follows the same pattern as `build_docker_vllm.sh`:

- Dynamically finds xDiT install via `python -c "import xfuser, os; ..."`
- For v26.6: checks out a specific commit before applying the patch
- For v26.7: applies patch directly to the image's xDiT
- Applies patch with `git apply || patch -p1 --fuzz=10` fallback
- Installs TraceLens via `pip install --upgrade`

### Merge (`trace_capture_merge_experimental.py`)

Extended `merge_capture_trace_into_graph()` with a `single_capture_trace` flag for diffusion traces. The key difference from LLM inference is the absence of multiple graph variants. For vLLM/SGLang, separate CUDA graphs are captured for each batch size, with `execution_details.json` mapping each capture trace to its batch size. For diffusion, a single graph is generated for the given resolution, and every denoise step replays that same graph. This means one capture trace is applied to all graph launches in the replay trace:

| | vLLM/SGLang | xDiT (diffusion) |
|---|---|---|
| Capture folder | Multiple files + `execution_details.json` | Single `.json.gz`, no metadata |
| Variant matching | Batch size → `find_closest_batch_size()` | N/A (single variant per run) |
| Graph root pairing | 1:1 (each execution root has its own graph launches) | 1:N (same capture root applied to all graph launches) |
| Execution roots | `execute_context_*` annotations | None |

The `single_capture_trace` flag is passed into `merge_capture_trace_into_graph()` and signifies that the single capture root will be applied to all graph launches.

### Report Generation (`generate_perf_report_pytorch_inference.py`)

Calls `merge_capture_trace_into_graph()` with `single_capture_trace` if the metadata json is empty.

### ROCTracer `wait=0` Bug
`torch.profiler.schedule(wait=0)` produces empty replay traces.
Workaround: `--profile_wait 1` via `EXTRA_XDIT_ARGS`.

### LTX graph capture profiling errors
Tracked here https://github.com/huggingface/diffusers/issues/14411. Prevents models like LTX that set `shift_terminal` from having graph capture profiling support.

### Categorizer Gaps
- `aten::mm->Cijk_*` synthetic op naming: categorizer doesn't recognize as GEMM. These ops have full `Input Dims` and are correctly analyzed by the perf model, but appear as "other" in category summaries.
- `aiter::fmha_v3_fwd`: AITER flash attention has no perf model.

### Tests (`test_inference_perf_report.py`)

- Added auto-detection of single capture trace mode (diffusion) vs multi-capture mode (vLLM/SGLang) in the test's capture merge logic, matching the inference report script.
- Added two xDiT regression test cases (`xdit_flux_aiter`, `xdit_flux_aiter_sage`) with reference traces and CSVs under `tests/traces/inference/`.
- Added `test_xdit_capture_merge` which validates the capture merge end-to-end for a specific GEMM shape (M=4608, K=15360, N=3072):
  1. Finds every `aten::mm` with the target dims in the capture trace
  2. Maps each to its child kernel's positional index in the capture root
  3. Uses those same indices into each `hipGraphLaunch` replay to find the exact replay kernels
  4. Asserts kernel names match between capture and replay at each index
  5. Sums replay kernel durations and asserts they match the report's `Kernel Time (µs)_sum`
  6. Asserts kernel count matches the report's `name_count`
  7. Verifies all kernel names from `kernel_details_summary` appear in `call_stack_full`, confirming that the capture merge properly populated call stack and kernel information.

## Files Changed

| File | Change |
|------|--------|
| `TraceLens/Trace2Tree/trace_capture_merge_experimental.py` | Add `single_capture_trace` flag, fallback graph root collection, auto-discover capture files |
| `tests/test_inference_perf_report.py` | Add single capture trace detection and xDiT regression tests |
| `tests/traces/inference/xdit_flux_aiter*` | Regression test traces and reference CSVs for xDiT |
| `TraceLens/Reporting/generate_perf_report_pytorch_inference.py` | Use `single_capture_trace` flag |
| `examples/.../build_docker_xdit.sh` | Build script supporting v26.6 and v26.7 |
| `examples/.../xdit_patches/config_xdit_v26.6.patch` | Patch for xDiT v26.6 |
| `examples/.../xdit_patches/config_xdit_v26.7.patch` | Patch for xDiT v26.7 |


TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -664,33 +671,56 @@ def find_execution_details(execution_root) -> Optional[str]:
 
 
 def merge_capture_trace_into_graph(
-    capture_folder: str,
-    metadata_json_path: str,
-    graph_tree_filepath: str,
+    capture_folder: str = None,
+    metadata_json_path: str = None,
+    graph_tree_filepath: str = None,
+    single_capture_trace: bool = False,
```
devalshahamd: Hi @kyle-hoffmeyer  Do we need this additional path if the user puts the capture trace in a folder and passes that? I think to make it future proof (where we might have multiple capture traces for a different model), it is better to keep it in a folder.

examples/custom_workflows/inference_analysis/xdit_patches/config_xdit_v26.7.patch
```diff
@@ -0,0 +1,120 @@
...
++            self._run_timed_pipe(compile_args)
++            return
++
++        # Phase 1: Dynamo trace + Inductor compile + eager warmup (1 step).
++        compile_args["num_inference_steps"] = 1
++        self._run_timed_pipe(compile_args)
++
++        # Phase 2: CUDA graph recording under profiler (1 step).
++        log("Profiling CUDA graph recording phase for shape trace...")
++        capture_dir = f"{self.config.output_directory}/capture_traces"
```
devalshahamd, line 90: We already have capture traces in a separate folder so we can pass the folder to the perf report script.

devalshahamd (commented):
Added some comments, there is also lint error

kyle-hoffmeyer:
Added stable-diffusion and hunyuanvideo tests

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
kyle-hoffmeyer: Even when there's a single capture trace, it still goes into a capture folder. The `single_capture_trace` var is just for `merge_capture_trace_into_graph` to know that there is a single capture trace and therefore no metadata JSON to help match capture roots with graph roots. Although maybe it'd be simpler if instead of passing this var into `merge_capture_trace_into_graph`, we detect within `merge_capture_trace_into_graph` whether there's a single capture trace (so we'd define `single_capture_trace` within `merge_capture_trace_into_graph`)

This would also remove diffs in generate_perf_report_pytorch_inference.py.

examples/custom_workflows/inference_analysis/xdit_patches/config_xdit_v26.7.patch
kyle-hoffmeyer: Yes the perf report script invocation is still the same. We pass the capture trace folder into generate_perf_report_pytorch_inference.py

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
devalshahamd: Hmm, I feel this part of reusing the same capture trace if there is only one is implicit. This implied way might be erroneous when the user provides a wrong capture folder or gives a capture folder with one trace by mistake (maybe the folder is not complete)...  Can we think about a more explicit approach for this? Maybe adding a CLI that says --single-capture-trace (not ideal since we are adding an argument).

Another approach is that in xDiT, we add an annotation to the profiler trace about what is the batch size so we can use the similar matching stratey-

1) Infer batch size/compilation model for a capture trace and store it in the execute_details.json file. We do not need to add annotation here since we can infer the batch size from shapes. 

2) Go through each execution step of  the graph replay, get batch size from annotation and number of hipgraphlaunches and then find the matching capture trace 

3) Merge the pair.

tests/traces/inference/hunyuanvideo/xdit_stdout.log
tsrikris, line 1: Making sure this one's needed

tests/traces/inference/xdit_sd_3.5/xdit_stdout.log
tsrikris, line 1: Similar here

tsrikris (approved):

kyle-hoffmeyer:
@devalshahamd,
- Modified patch to generate annotation for replay phase
- Annotation is for every full denoising workload
- Removed `single_capture_trace` logic. Relies on vllm/sglang annotation matching approach
- Old logic for inferring piecewise/full mode from execution root no longer holds
  - Old logic assumed that each execution root corresponded to a single iteration and thus should contain one graph launch if in full mode
  - However, since each execution root is for a full denoising workload, it is bound to contain multiple graph launches for each denoising step
- For a given batch size we now check `execution_details.json` to infer mode from capture trace.
  - If only a full mode trace exists, assume full (likely xDiT)
  - If only a piecewise mode trace exists, assume piecewise (likely xDiT)
  - If both exist, use old logic (likely vllm/sglang)

See `tests/traces/inference/flux` for an example. If this looks good, I will update the rest of the references which should fix failing tests.

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
```diff
@@ -718,26 +720,49 @@ def merge_capture_trace_into_graph(
...
+        # capture root (e.g. diffusion where one captured graph is replayed
+        # across many denoising steps), map it to every graph root.
+        # Otherwise pair 1:1 positionally (vLLM/SGLang).
+        if len(capture_roots) == 1:
+            pairs = [(capture_roots[0], g) for g in graph_roots]
+            data = [capture_root_data[0]] * len(graph_roots)
+        else:
+            pairs = list(zip(capture_roots, graph_roots))
+            data = capture_root_data
+
```
devalshahamd: We can reduce the docstring added. I also realized that there is a silent bug here that when number of capture_roots!=number of graph_roots, the zip() functional will silently drop it. We can add a message in the else condition ("warning: number of capture and graph roots differ, merging might be erroneous")

devalshahamd: I don't want to fix it as of now, since this will be covered in the fallback path work. Just a warning message is good.

devalshahamd (changes requested):
LGTM! You can remove the old perf reports and add a warning message for my minor comment. Let me know when you are done so I can approve. You might need to do rebase!

TraceLens/Trace2Tree/trace_capture_merge_experimental.py
gabeweisz: The PR needs to be updated anyway because tests failed. Any reason not to fix this when fixing them?

devalshahamd: We never had this check "len(capture_roots)==len(graph_roots)" so I am afraid adding a hard check now might result in failures in some cases and we can add a check as we implement aspects of the fallback path plan. However, thinking about it more, I don't think hard check will be an issue and is good to add that check.

We can have elif len(capture_roots)==len(graph_roots) instead of else, and then have a warning message in "else"

devalshahamd (approved):
LGTM, thanks!

PR #849, opened 2026-07-27
Update check for local server readiness

gabeweisz:
Don't rely on curl to see if the local server is ready


Ahmedhasssan-aig (approved):
LGTM!

PR #858, opened 2026-07-30
Feat/gw fix auto evals

gabeweisz:
Fix generate refs script to allow local agent
Fix parallel eval script to have explicit "both"


gabeweisz:
Running the test now, hold this until it finishes

PR #869, opened 2026-07-31
Feat/gw fix cyclic imports

gabeweisz:
Goes on top of https://github.com/AMD-AGI/TraceLens/pull/868

PR #867, opened 2026-07-31
Remove unused imports across the codebase

gabeweisz:
Remove unused imports across the codebase

PR #870, opened 2026-07-31
Fix signature mismatches and missing parent class init calls

### LINKED ISSUE #875: Fix signature mismatches in inheritance

### LINKED ISSUE #879: Fix missing call to parent class init

### LINKED ISSUE #880: Fix incorrect signature for override function

### LINKED ISSUE #883: Fix calls with the wrong arguments

### LINKED ISSUE #884: Fix mixed returns

gabeweisz:
Fixes #875
Fixes #879
Fixes #880
Fixes #883
Fixes #884 

## Summary
- Align override signatures with base classes in perf models, tree perf, GPU analyser, and JAX/PyTorch tree utilities so subclasses remain LSP-compatible (#875)
- Call `super().__init__()` in MoE, quant, and attention perf-model subclasses that were duplicating base initialization instead of delegating to the parent (#879)
- Includes prerequisite cleanup from the stacked lint/import PRs (#867–#869) this branch builds on

## Test plan
- [x] Black, Ruff (F401/F841), and import-linter pass on changed files
- [x] CI green on this PR

Comes after #869


tsrikris (approved):

devalshahamd (commented):
I think this has changes from the PR 869 as well. So comments from that PR apply here as well

TraceLens/PerfModel/extensions/perf_model_extensions.py
```diff
@@ -475,20 +475,8 @@ class GroupQuant(BinaryElementwise):
     category = "GroupQuant"
     bwd_category = None
 
-    def __init__(self, event, arch=None, python_path=None):
```
devalshahamd, line 478: Needs its own __init__ because the second input is a scales tensor rather than a
    broadcast operand. Its trailing dim is ceil(N / group_size), which is not
    broadcastable against N, so the parent's derived output shape raises ValueError.
    The output shape and dtype are read directly from the trace instead; the latter
    also avoids the parent's max(bpe_in1, bpe_in2) fallback, which is backwards for
    quantization since the output is narrower than the input.

ajassani:
**Blocking:** The new `super().__init__()` call breaks `vllm_triton_per_token_group_quant_fp8` construction.

Minimal reproducer:

```python
def test_vllm_group_quant_constructs_without_trace():
    event = {
        "args": {
            "Input Dims": ((4, 256), ()),
            "Input type": ("c10::BFloat16", "Scalar"),
            "Concrete Inputs": ("", "128"),
        }
    }

    model = vllm_triton_per_token_group_quant_fp8(event)
    assert model.param_details["M"] == 4
    assert model.param_details["N"] == 256
    assert model.param_details["group_size"] == 128
```

Here `(4, 256)` represents four rows/tokens with 256 features per row; group size 128 gives two quantization groups per row. The values are intentionally small and synthetic so the constructor contract can be tested without a trace.

This passes on parent #869 but fails on the latest #870 head with:

```text
KeyError: 'shape_in1'
```

`BinaryElementwise.__init__()` expects `shape_in1`/`shape_in2`, while this subclass returns `M`/`N`. Could we preserve its specialized initialization or align the parameter contract?

We should also add similar no-trace constructor tests for the other perf models changed by this PR, with basic `flops()` and `bytes()` assertions.

gabeweisz:
> I think this has changes from the PR 869 as well. So comments from that PR apply here as well

Yes, we need to pull main after merging 869 to get rid of duplicates

TraceLens/PerfModel/extensions/perf_model_extensions.py
gabeweisz: Fixing

gabeweisz:
> **Blocking:** The new `super().__init__()` call breaks `vllm_triton_per_token_group_quant_fp8` construction.
> 
> Minimal reproducer:
> 
> ```python
> def test_vllm_group_quant_constructs_without_trace():
>     event = {
>         "args": {
>             "Input Dims": ((4, 256), ()),
>             "Input type": ("c10::BFloat16", "Scalar"),
>             "Concrete Inputs": ("", "128"),
>         }
>     }
> 
>     model = vllm_triton_per_token_group_quant_fp8(event)
>     assert model.param_details["M"] == 4
>     assert model.param_details["N"] == 256
>     assert model.param_details["group_size"] == 128
> ```
> 
> Here `(4, 256)` represents four rows/tokens with 256 features per row; group size 128 gives two quantization groups per row. The values are intentionally small and synthetic so the constructor contract can be tested without a trace.
> 
> This passes on parent #869 but fails on the latest #870 head with:
> 
> ```
> KeyError: 'shape_in1'
> ```
> 
> `BinaryElementwise.__init__()` expects `shape_in1`/`shape_in2`, while this subclass returns `M`/`N`. Could we preserve its specialized initialization or align the parameter contract?
> 
> We should also add similar no-trace constructor tests for the other perf models changed by this PR, with basic `flops()` and `bytes()` assertions.

Fixing

ajassani:
Was this change intentional?

```diff
- return super().bytes_bwd(bpe)
+ return super().bytes(bpe)
```

It affects both backward convolution models and changes a representative result from 121 bytes to 60. The surrounding comment and `flops_bwd()` path suggest `bytes_bwd()` is intended. Is there a reason the forward byte formula is correct here?

gabeweisz:
> Was this change intentional?
> 
> ```diff
> - return super().bytes_bwd(bpe)
> + return super().bytes(bpe)
> ```
> 
> It affects both backward convolution models and changes a representative result from 121 bytes to 60. The surrounding comment and `flops_bwd()` path suggest `bytes_bwd()` is intended. Is there a reason the forward byte formula is correct here?

Will review. Cursor primarily authored the changes

gabeweisz:
> > Was this change intentional?
> > ```diff
> > - return super().bytes_bwd(bpe)
> > + return super().bytes(bpe)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It affects both backward convolution models and changes a representative result from 121 bytes to 60. The surrounding comment and `flops_bwd()` path suggest `bytes_bwd()` is intended. Is there a reason the forward byte formula is correct here?
> 
> Will review. Cursor primarily authored the changes

The change was wrong and has been reverted

ajassani (approved):

devalshahamd (approved):

spandoesai (approved):
LGTM!

PR #904, opened 2026-08-05
Tolerate sub-microsecond event overlap

### LINKED ISSUE #900: Tree building event overlap tolerance
## Summary

The tree builder's "event bleed" detection uses a strict `>` comparison to detect when a child event ends after its parent. Floating point rounding on large microsecond timestamps can produce sub-nanosecond differences between events that end at the same real time, causing true parent-child pairs to be misclassified as bleeds and discarded from the tree.

## Location

`TraceLens/Trace2Tree/trace_to_tree.py`, lines 163-165:

```python
if stack and (
    event[TraceLens.util.TraceEventUtils.TraceKeys.TimeEnd]
    > stack[-1][TraceLens.util.TraceEventUtils.TraceKeys.TimeEnd]
):
```

## What Happens

The tree builder processes events sorted by timestamp and maintains a stack to track the call hierarchy. When an event starts inside the current stack top but ends after it, the code treats this as "event bleed" — the event doesn't properly nest inside its parent.

The bleed handling has two stages:

1. **Detection** (line 163): `event.end > stack_top.end` — strict `>`, no tolerance.
2. **Handling** (lines 167-179): If the overlap (how much the event overlaps with the stack top) is < 1 us, the stack top is popped and the event is added as a sibling. If the overlap is >= 1 us, the event is discarded. This 1 us tolerance was added in PR #694 to handle sub-microsecond clock jitter between siblings.

The problem is that stage 1 has no tolerance. A 0.0005 us (500 picosecond) floating point artifact in the end-time comparison triggers stage 2, where the overlap is computed as the full duration of the event (hundreds of milliseconds) — far exceeding 1 us — so the event is discarded. The tolerance in stage 2 doesn't help here because it's designed for the sibling case (small overlap between adjacent events), not the parent-child case (child ends at the same real time as parent but floating point says otherwise).

## Specific Example

Two `python_function` events that end at the same real time (the PyTorch profiler records them with identical wall-clock end times):

| Event | `ts` (us) | `dur` (us) | `ts + dur` (computed) |
|---|---|---|---|
| `<frozen runpy>(88): _run_code` | 2987767959258.654 | 773660.778 | 2987768732919.431**6** |
| `/app/xDiT/xfuser/runner.py(126): <module>` | 2987767959260.416 | 773659.016 | 2987768732919.432**0** |

The real end time is the same, but double-precision arithmetic produces different results when adding different `ts` and `dur` pairs. The difference is 0.0005 us (500 picoseconds).

Because `runner.py` end (2987768732919.4320) > `_run_code` end (2987768732919.4316), the bleed check triggers. The overlap is then computed as:

```
overlap = _run_code.end - runner.ts = 2987768732919.4316 - 2987767959260.416 = 773659.016 us
```

This is 773 ms — far exceeding the 1 us tolerance — so `runner.py` is discarded. This cascades: `runner.py(81): profile` and `base_model.py: profile` are also discarded because they have the same end-time rounding issue. The call stack in the final report starts at `_run_timed_pipe` instead of `runner.py <module>`.

The same trace collected again happened to produce identical floating point results for `ts + dur`, so no bleed was detected and the tree was built correctly.

## Comparison

Running the instrumented tree builder on both traces:

**trace1** — no bleed, properly parented:
```
ADDED: <frozen runpy>(88): _run_code           parent=<frozen runpy>(198): _run_module_as_main
ADDED: runner.py(126): <module>                parent=<frozen runpy>(88): _run_code
ADDED: runner.py(81): profile                  parent=runner.py(126): <module>
ADDED: base_model.py: profile                  parent=runner.py(81): profile
```

**trace2** — false bleed, orphaned:
```
ADDED:   <frozen runpy>(88): _run_code         parent=<frozen runpy>(198): _run_module_as_main
SKIPPED: runner.py(126): <module>              overlap=773659.016us (bleed)
SKIPPED: runner.py(81): profile                overlap=773658.054us (bleed)
SKIPPED: base_model.py: profile                overlap=773653.868us (bleed)
ADDED:   _run_timed_pipe                       parent=<frozen runpy>(88): _run_code
```

## Suggested Fix

Add a numerical tolerance to the bleed detection comparison.

```python
bleed_tolerance_us = x  # ex. 1 nanosecond
if stack and (
    event[TE] > stack[-1][TE] + bleed_tolerance_us
):
```

The specific tolerance will need to be tested across a set of traces to ensure this doesn't incorrectly assign events as children.

kyle-hoffmeyer:
## Summary

The tree builder's "event bleed" detection in `build_host_call_stack_tree` uses a strict `>` comparison to decide when an event ends after the current stack top. Due to profiling noise and floating point rounding there can be sub-nanosecond time differences between two events that really end at the same wall-clock time. When that happens, the strict comparison misclassifies a true parent-child pair as a bleed and discards the child.

The existing bleed-handling logic (added in #694) handles time jitter between siblings, but not for time jitter between parent-children pairs.

See #900: `runner.py(126): <module>` and its descendants (real `python_function` events that end at the same wall-clock time as their parent) were dropped from the tree in one run of a trace but not an identical rerun, purely because `ts + dur` rounded to a slightly different value depending on operand magnitudes.

### Solution
Apply the same 1us tolerance already used for handling bleeds to the detection comparison itself: `event.end > stack_top.end + tolerance`. An event that ends within 1us of its parent's end is no longer treated as a bleed.

## Implementation

- `TraceLens/Trace2Tree/trace_to_tree.py`: moved the `overlap_tolerance_us = 1.0` constant above the bleed-detection `if` and added it to the detection comparison's right-hand side.

## Tests

- Verified against the exact scenario from #900 using local xDiT hunyuanvideo trace. The `call_stack_full` column in `unified_perf_summary` shows the fix restoring the missing top-level frames:

**Pre-fix** — `runner.py(126): <module>` is dropped as a false bleed, so the call stack starts one level too deep:
```
runner.py(81): profile
  └─ base_model.py(611): profile
       └─ base_model.py(677): _run_timed_pipe
            └─ hunyuan.py(79): _run_pipe
                 └─ ...
```

**Post-fix** — the same event is correctly nested as a child, restoring the full call stack:
```
runner.py(126): <module>        ← restored
  └─ runner.py(81): profile
       └─ base_model.py(611): profile
            └─ base_model.py(677): _run_timed_pipe
                 └─ hunyuan.py(79): _run_pipe
                      └─ ...
```

Closes #900


spandoesai (approved):
LGTM! Retains past behaviour along with new fixes

gabeweisz (approved):

devalshahamd (approved):

PR #905, opened 2026-08-06
Enable Codecov, CodeQL, and coverage regression gate

gabeweisz:
## Summary
- Enforce Codecov upload in CI (`fail_ci_if_error`)
- **Block PRs when overall project coverage drops** (`codecov/project`: `target: auto`, `threshold: 0%`)
- **Block PRs when patch coverage on changed lines is insufficient** (`codecov/patch`: `target: auto`, blocking)
- GitHub Checks annotations on coverage results
- Add **CodeQL** security scanning for Python on push/PR to main (weekly schedule)
- **Upload JUnit test results to Codecov Test Analytics** (flag: `unit`) so skipped GPU tests are visible with skip reasons in CI
- Configure **`unit` and `gpu` Codecov flags**; GPU-only paths (`EventReplay/`, `PerfModel/benchmarking/`) use **carryforward** when GPU CI is absent
- Register **`@pytest.mark.gpu`** in `tests/conftest.py` and mark GPU integration tests (includes #916)

Follow-up to #898. Closes #894. Supersedes #916 (merged here). Current baseline is ~50% project coverage ([comment](https://github.com/AMD-AGI/TraceLens/pull/898#issuecomment-5197227856)).

## What is enforced
| Check | Enforced? |
|-------|-----------|
| Coverage upload succeeds in CI | Yes |
| JUnit test results upload succeeds | Yes |
| `codecov/project` — no coverage regression | Yes (blocking) |
| `codecov/patch` — changed lines covered | Yes (blocking) |
| GitHub Checks annotations | Yes |
| CodeQL findings | Yes (blocking) |
| GPU coverage carryforward | Informational (`gpu` flag) |

## Test analytics & GPU skips
- CI emits `junit.xml` alongside `coverage.xml`
- Skipped `@pytest.mark.gpu` tests appear in Codecov Test Analytics with skip reasons (e.g. no torch/GPU in CPU CI)
- Morning test PR branches (#909–#915) already merge this infra; #914/#915 add GPU markers on new test classes

## Test plan
- [x] Codecov CI job passes and uploads coverage + JUnit
- [x] `codecov/project` fails if coverage drops vs base
- [x] `codecov/patch` fails if changed lines lack adequate coverage
- [x] PR comment shows project + patch coverage and carryforward flags
- [x] CodeQL workflow completes

After https://github.com/AMD-AGI/TraceLens/pull/903 and https://github.com/AMD-AGI/TraceLens/pull/902

PR #914, opened 2026-08-07
Add unit tests for PerfModel benchmarking

gabeweisz:
## Summary
- Add `tests/test_perfmodel_benchmarking.py` with 33 tests for `TraceLens/PerfModel/benchmarking/`.
- CPU-only coverage runs in CI without torch: `microbench_utils` device mapping, metric coercion, and idle-check fallbacks (mocked nvidia-smi path).
- Torch-dependent modules are imported lazily; classes are skipped when torch is not installed.
- GPU benchmark smoke tests (`bench_gemm`, MXFP4 tensor prep, rocprof single-run) require CUDA/HIP and skip otherwise.

## Test plan
- [x] `python -m pytest tests/test_perfmodel_benchmarking.py -v` (12 passed, 21 skipped locally without torch/GPU)
- [x] Black, Ruff (F401/F841/RET502/RET503), and import-linter pass on changed files

After #905 

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -37,13 +37,23 @@ jobs:
       - name: Run unit tests
         run: |
           . .venv/bin/activate
-          python -m pytest tests/ -v --tb=short -n auto --dist loadgroup --cov=./TraceLens --cov-report=xml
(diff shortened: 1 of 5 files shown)


Ahmedhasssan-aig (approved):
LGTM!

PR #920, opened 2026-08-07
Add unit tests for TraceLens.TreePerf

gabeweisz:
## Summary
- Add `tests/test_treeperf.py` covering `GPUEventAnalyser`, `JaxAnalyses`, `TreePerfAnalyzer`, and `JaxTreePerfAnalyzer` with synthetic CPU-safe traces.
- Include optional `@pytest.mark.gpu` live-profile tests (skipped in CPU-only CI via `importorskip` + CUDA check).

## Test plan
- [x] `pytest tests/test_treeperf.py -m "not gpu"` (48 passed)
- [x] black and ruff pass
- [x] CI unit tests pass

After #905


ajassani:
`test_gpu_event_analyser_compute_metrics_on_live_trace` is always skipped in the current unit-test job because PyTorch is not installed, so it contributes no CI coverage. Since the repository already has committed trace-based tests for the analyzer path, can we remove this test? If we want to preserve the current-PyTorch-profiler compatibility check, it would be more useful in a dedicated GPU integration suite that installs PyTorch and actually runs it.

gabeweisz:
I was adding so it is ready when we actually do have GPUs for CI. But if you want we can remove it - it might interfere with the codecov

ajassani:
Agreed—let’s keep it for future GPU CI. It shouldn’t materially affect Codecov because TraceLens post-processes profiler output, and the stored-trace tests already cover the same analyzer paths. When GPU CI arrives, we should centralize live-trace generation rather than duplicate profiler setup here; EventReplay already has the `profile_resnet` trace-generation function that we can reuse or generalize.

tests/test_treeperf.py
```diff
@@ -0,0 +1,761 @@
...
+            model(x).sum().backward()
+            prof.step()
+
+    analyzer = TreePerfAnalyzer.from_file(str(trace_path), rebuild_tree=True)
+    assert analyzer.check_gpu_only() is False
+    assert "busy_time" in analyzer.get_df_gpu_timeline()["type"].values
+    assert len(analyzer.get_kernel_launchers()) >= 1
+
+
+@pytest.mark.gpu
```
ajassani: We need a stronger test for `build_nn_module_latency_tree`. This test is accidentally skipped in CI because it is GPU-gated despite not using a GPU. Could we replace it with a CPU test using the existing SGLang prefill/decode trace? For example:

```python
def test_build_nn_module_latency_tree():
    trace_path = os.path.join(
        os.path.dirname(__file__),
        "traces/inference/sglang_prefilldecode/"
        "sglang_Qwen3-8B_prefilldecode.json.gz",
    )
    analyzer = TreePerfAnalyzer.from_file(
        trace_path, add_python_func=True
    )
    tree = analyzer.tree
    root = next(
        e for e in tree.events
        if e.get("name") == "nn.Module: Qwen3Model_0"
    )

    analyzer.build_nn_module_latency_tree(root)
    children = [
        tree.get_UID2event(uid)
        for uid in tree.get_nn_module_children(root)
    ]

    assert sum(
        child["name"].startswith("nn.Module: Qwen3DecoderLayer_")
        for child in children
    ) == 36
    assert root["GPU Time"] == pytest.approx(492163.76953125)
```

This runs CPU-only and validates a real nested module hierarchy. I tested it locally against the committed trace.

gabeweisz:
> Agreed—let’s keep it for future GPU CI. It shouldn’t materially affect Codecov because TraceLens post-processes profiler output, and the stored-trace tests already cover the same analyzer paths. When GPU CI arrives, we should centralize live-trace generation rather than duplicate profiler setup here; EventReplay already has the `profile_resnet` trace-generation function that we can reuse or generalize.

I already removed it but we can recreate later

tests/test_treeperf.py
gabeweisz: Done, please review again

ajassani: Thanks, the real CPU trace test is much better. Could we add assertions for one nested module to verify the recursive outputs, not only the root? For example:

```python
layer = next(
    child for child in children
    if child["name"] == "nn.Module: Qwen3DecoderLayer_0"
)
attention = next(
    tree.get_UID2event(uid)
    for uid in tree.get_nn_module_children(layer)
    if tree.get_UID2event(uid)["name"] == "nn.Module: Qwen3Attention_0"
)

assert layer["nn Parent GPU Time"] == pytest.approx(root["GPU Time"])
assert attention["nn Parent GPU Time"] == pytest.approx(layer["GPU Time"])
assert attention["Non-nn.Module GPU Time"] > 0
```

This ensures recursion and the non-module attribution branch are actually validated.

ajassani (approved):

PR #924, opened 2026-08-07
Add perf model for aten::replication_pad (1d/2d/3d)

kyle-hoffmeyer:
### Summary

TraceLens had no performance model for `aten::replication_pad3d` (or its 1d/2d variants). In diffusion workloads — particularly HunyuanVideo running through xDiT — replication padding accounts for **5.3% of GPU runtime** across 18 op instances, all in the VAE decoder's `HunyuanVideoCausalConv3d` layer. Without a perf model, these ops were bucketed as "other" with no roofline analysis, leaving a meaningful gap in performance coverage for video diffusion models.

#### What replication padding does

Replication padding extends a tensor's spatial dimensions by copying edge values outward. For a 1D example with input `[A, B, C, D]` and padding `[2, 1]` (2 left, 1 right):

```
Output: [A, A, A, B, C, D, D]
         ^  ^  ^  ^  ^  ^  ^
         replicated  original  replicated
         from edge   content   from edge
```

Every input element is copied to the interior of the output. Border positions replicate the nearest edge value.

#### Why it exists in HunyuanVideo

HunyuanVideo's VAE decoder uses causal 3D convolutions (`HunyuanVideoCausalConv3d`). A 3D conv with kernel size 3 along the temporal axis would shrink 17 frames to 15. Replication padding adds elements to preserve the output size. The padding is asymmetric — `[1, 1, 1, 1, 2, 0]` meaning 1 left/1 right (width), 1 top/1 bottom (height), 2 front/0 back (temporal) — so that each output frame depends only on the current and past input frames (causal constraint). No back padding is needed because the convolution already has enough frames to process the end of the sequence.

### Implementation

**`TraceLens/PerfModel/perf_model.py`** — adds `aten_replication_pad`, a new perf model class inheriting from `UnaryElementwise`.

`get_param_details()` parses the input tensor shape and the padding list from `Concrete Inputs` (e.g. `[1, 1, 1, 1, 2, 0]` for 3d left/right/top/bottom/front/back), then computes the output shape by adding each padding pair to the corresponding spatial dimension. The padding list is ordered innermost-dim-first, matching PyTorch's convention.

#### Bytes calculation

The kernel launches one thread per output element. Each thread computes its output coordinate, clamps the spatial indices to the valid input range (this is the "replication" — out-of-bounds coordinates snap to the nearest edge), issues one global memory load to read the input element, and one global memory store to write the output element. There is no arithmetic beyond integer address computation.

The perf model counts unique HBM data movement:

- **Reads**: every input element is read at least once (the full input is copied to the output's interior), so `bytes_read = nelems_in × bpe`.
- **Writes**: every output element is written exactly once, so `bytes_write = nelems_out × bpe`.
- **Total**: `bytes = nelems_in × bpe + nelems_out × bpe`.

For the HunyuanVideo example — input `(1, 128, 17, 240, 240)` bf16, padding `[1,1,1,1,2,0]`, output `(1, 128, 19, 242, 242)`:

```
nelems_in  = 1 × 128 × 17 × 240 × 240 = 125,337,600   → reads  = 125.3M × 2 = 250.7 MB
nelems_out = 1 × 128 × 19 × 242 × 242 = 142,427,648   → writes = 142.4M × 2 = 284.9 MB
                                                          total  = 535.5 MB
```

Note: the kernel actually issues `nelems_out` (142.4M) load instructions, not `nelems_in` (125.3M), because border threads re-read the same edge element as their neighbor. The perf model counts `nelems_in` for the read side because those duplicate border should hit L2 cache and not generate additional HBM traffic.

#### FLOPs calculation

`flops()` is inherited from the `UnaryElementwise` base class as `nelems_out` (142.4M). This represents one copy operation per output element — there is no multiply, add, or activation. The FLOPS/Byte ratio of ~0.27 correctly classifies the op as memory-bound on the roofline.

The class is placed right after `aten_unary_elementwise` since it shares the same base class and follows the same pattern.

**`TraceLens/PerfModel/torch_op_mapping.py`** — registers all three variants (`aten::replication_pad1d`, `pad2d`, `pad3d`) to the new class, following the same loop pattern used for `unary_elemwise_ops`.

### Tests

- Synthetic event tests verified output shape computation (including zero-padding and 2d cases) and correct bytes/FLOPS values against manual calculation.
- Hardware counter validation via `rocprofv3 --pmc` on MI300X using the `validate-perf-model` harness with the HunyuanVideo trace shape `(1, 128, 17, 240, 240)` + padding `[1,1,1,1,2,0]`:

  | Metric | Predicted | Measured (rocprofv3) | Ratio |
  |---|---|---|---|
  | HBM bytes (read + write) | 535.5 MB | 651.3 MB | 1.22x |
  | MFMA ops (I8 + F8) | 0 | 0 | — |
  | Achieved HBM BW | 1.13 TB/s | 1.38 TB/s | — |

  The 1.22x bytes ratio is within expected range — the overshoot comes from L2 cache-line granularity (128B lines) and TCC read request replays for non-aligned accesses. Zero MFMA ops confirms the op is pure bandwidth-bound.

- Full regression suite: 1467 passed, 0 new failures introduced (all 7 pre-existing failures are unrelated).


tsrikris (approved):

Ahmedhasssan-aig:
Please resolve the conflicts

ajassani (approved):

devalshahamd (approved):
LGTM!

PR #926, opened 2026-08-07
Add perf model for aiter::_groupnorm_run

kyle-hoffmeyer:
### Summary

`aiter::_groupnorm_run` is AITER's optimized group normalization kernel, used in xDiT diffusion workloads. It had no perf model in TraceLens, so it was bucketed as "other" with no roofline analysis. It accounts for 1.35% of GPU runtime in HunyuanVideo (14 instances) and ~0.5% in each Flux trace.

#### What the operation does

Group normalization divides a tensor's channels into groups and normalizes each group independently. For input `(1, 128, 17, 240, 240)` with 32 groups:

- 128 channels / 32 groups = 4 channels per group
- Each group has `4 x 17 x 240 x 240 = 3,916,800` elements

For each group, the kernel:
1. Computes the mean across all elements in the group
2. Computes the variance (sum of squared differences from mean)
3. Normalizes each element: `(x - mean) / sqrt(variance + eps)`
4. Applies a learned per-channel affine transform: `y = normalized * weight + bias`

The AITER implementation (`aiter/ops/groupnorm.py` in [ROCm/aiter](https://github.com/ROCm/aiter)) is a JIT-compiled two-pass HIP kernel. The first pass (`groupnorm_kernel_dispatch_up`) computes per-group mean and variance using a reduction into a small workspace buffer (~32 KB, cached across calls). The second pass (`groupnorm_kernel_dispatch_down`) normalizes and applies the affine transform. This is functionally identical to PyTorch's `aten::group_norm` but uses a fused two-kernel approach instead of PyTorch's three-kernel implementation.

#### Flops/bytes calculation

The existing `GroupNorm` class (inheriting from `Normalization` in `perf_model.py`) models this as:

**Bytes:**
Same as `GroupNorm`

```
activation_bytes = nelems * bpe_in + nelems * bpe_out    (read input + write output)
weight_bytes     = num_weight_tensors * C * bpe_in       (read weight, bias, mean, var)
total            = activation_bytes + weight_bytes
```
For the HunyuanVideo shape `(1, 128, 17, 240, 240)` bf16: `nelems = 125,337,600`, `bpe = 2`, giving ~501 MB total. The weight tensors (mean, variance, scale, bias — 4 tensors of 128 floats each) add negligible bytes.

**FLOPs:**
Same as `GroupNorm`

```
param_compute  = C * 3                    (per-channel: 1/std * mean, alpha, beta)
               + C * 3                    (per-channel: inverse sqrt of var+eps)
               + nelems * 3               (training: sum for mean, subtract mean, sum squared diffs)
actual_compute = 2 * nelems               (normalize: multiply by alpha, add beta)
total          = param_compute + actual_compute ≈ 5 * nelems
```
Giving ~627M FLOPs. The FLOPS/Byte ratio of ~1.25 places the op firmly on the memory-bound side of the roofline.

### Implementation

**`TraceLens/PerfModel/extensions/pseudo_ops_perf_utils.py`** — one-line registration mapping `"aiter::_groupnorm_run"` to the existing `GroupNorm` class. No new class was needed because the AITER trace event has the same `Input Dims` layout as `aten::group_norm` — input shape at `[0]`, weight at `[2]`, bias at `[3]`, with `num_groups` in `Concrete Inputs[1]`. The existing `GroupNorm.get_param_details()` parses this correctly.

The registration goes in `pseudo_ops_perf_utils.py` (not `torch_op_mapping.py`) because `aiter::` ops are vendor-specific pseudo-ops, not standard PyTorch `aten::` operators.

### Tests

- Hardware counter validation via `rocprofv3 --pmc` on MI300X with HunyuanVideo shape `(1, 128, 17, 240, 240)`, 32 groups:

  | Metric | Predicted | Measured (rocprofv3) | Ratio |
  |---|---|---|---|
  | HBM bytes (read + write) | 501.35 MB | 501.47 MB | 1.0002x |
  | MFMA ops (I8 + F8) | 0 | 0 | — |
  | Achieved HBM BW | 2.62 TB/s | 2.62 TB/s | — |

  The bytes prediction is essentially a perfect match (0.02% error), confirming the model accurately captures the kernel's memory traffic. Zero MFMA ops confirms the op is bandwidth-bound — all compute is scalar VALU work.

- HunyuanVideo inference perf report reference CSVs updated: all 14 `aiter::_groupnorm_run` instances now show `has_perf_model=True` with `category=NORM_fwd`.
- Full regression suite: 1467 passed, 0 new failures.


devalshahamd (approved):
LGTM!

PR #943, opened 2026-08-14
feat(arch): add benchmark-derived Radeon8060S (Strix Halo) arch spec

bong-water-water-bong:
## What

Adds `arch/Radeon8060S.json` — a benchmark-derived architecture spec for the AMD Radeon 8060S (Strix Halo APU, gfx1151), plus a README entry documenting the source.

## Method

Numbers measured on target hardware with the existing microbench harness:
`python -m TraceLens.PerfModel.benchmarking.microbench --device 0 --output <name>.json`
— merged **max of 3 runs** (roofline semantics). Key values:

| Metric | Value |
|---|---|
| matrix bf16 / fp16 / int8 | 36 / 37 / 37 TFLOPS |
| matrix fp32 | 3 TFLOPS |
| fp8 | 0 (no path on ROCm 10.1 stack) |
| fp4 / fp6 | 5 TFLOPS |
| HBM read / write | 202 / 231 GB/s |
| memory | 120 GB unified (LPDDR5X) |

## Why it matters

When profiling this very common APU, users previously had to fall back to an Instinct proxy arch — which was **20–25× off** the real roofline (e.g. MI300X's 1300+ TFLOPS vs the measured 36). With a real 8060S arch, TraceLens analysis on Strix Halo targets produces quantified headroom instead of unmodeled-heuristic findings.

Note: thermal band documented in the README — matrix ops swing 9–21% between runs.

### DIFF
--- a/TraceLens/Agent/Analysis/utils/arch/README.md
+++ b/TraceLens/Agent/Analysis/utils/arch/README.md
@@ -9,6 +9,11 @@ These performance metrics are derived from the following sources:
 
 - [Understanding Peak and Max-Achievable FLOPS](https://rocm.blogs.amd.com/software-tools-optimization/Understanding_Peak_and_Max-Achievable_FLOPS/README.html)
 - [AMD Instinct MI300X](https://www.amd.com/en/products/accelerators/instinct/mi300/mi300x.html)
+- [AMD Instinct MI300X](https://www.amd.com/en/products/accelerators/instinct/mi300/mi300x.html)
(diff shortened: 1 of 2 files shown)


tsrikris:
Thanks for submitting a PR! The recommendation for MI355X and beyond is to use a benchmarking [script](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/PerfModel/benchmarking/README.md) as we'd indicated in the TraceLens Agent [README](https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/Agent/Analysis/README.md)

PR #948, opened 2026-08-17
Fix xDiT replay profiling

kyle-hoffmeyer:
The graph replay profiling was broken and the patch was not properly patching the replay section.

Graph replay profiling targets the `profile()` method in xDiT. When `profile()` is called, it runs and profiles the actual workload (replay phase).
The profiler has a default annotation. When `self.config.profile_capture_phase` is true, this default annotation is substituted with the TraceLens annotation.


devalshahamd (approved):
LGTM!

PR #952, opened 2026-08-18
feat(vllm): instrument the V2 model runner for graph-capture traces

tkarttunen:
## Summary

The graph-capture profiler is wired into `capture_model()` in the legacy `GPUModelRunner` only. Since vLLM 0.22, `VllmConfig.use_v2_model_runner` selects the rewritten runner under `vllm/v1/worker/gpu/` by default for many configurations, and that runner contains no reference to `capture_torch_profiler` at all.

The failure is silent. The patch applies, the server starts and serves, CUDA graphs are captured normally, `capture_traces/` stays empty, and neither of the patch's own log lines is emitted. Nothing reports an error, so the only way to notice is to go looking for traces that were never written. The workaround has been to pin `VLLM_USE_V2_MODEL_RUNNER=0`, which forces users onto a runner vLLM no longer selects by default.

This instruments `CudaGraphManager.capture()` in `vllm/v1/worker/gpu/cudagraph_utils.py`, which owns the per-shape capture loop for both PIECEWISE and FULL modes, for 0.22 through 0.25.

## What changed

- `CudaGraphManager` already holds `vllm_config`, so the profiler is built where it is used. Nothing is threaded through `model_runner.py` the way the legacy patch threads a `profiler=` kwarg down two call levels.
- Behaviour matches the legacy path exactly: same rank-0 guard, same trace directory (`torch_profiler_dir/capture_traces`), same `worker_name`, and the same `capture_{num_tokens}_{mode}` annotation that `CaptureAnnotation`'s `capture_(\d+)_(.*)` pattern parses.
- Warmup passes stay outside the profiled region, matching the legacy runner.
- Additions only: 514 lines across the four patch files, no deletions.

## Evidence

Verified on MI350X for 0.22 and 0.25, the two ends of the affected range, with no `VLLM_USE_V2_MODEL_RUNNER` set anywhere:

```
INFO [gpu_worker.py] Using V2 Model Runner
INFO [cudagraph_utils.py] Rank 0: Torch profiler enabled for CUDA graph capture,
     traces will be saved to: /tmp/traces/capture_traces
```

| Version | Runner selected | Capture shards | Annotation sample |
|---|---|---|---|
| 0.22 | V2 | 102 | `capture_512_PIECEWISE` |
| 0.25 | V2 | 102 | `capture_512_PIECEWISE` |

102 shards is the same count the legacy runner produces for the same model, so the V2 path now yields equivalent output rather than merely non-empty output.

## Test plan

- [x] All four patches apply to their base images (`vllm/vllm-openai-rocm:v0.2{2,3,4,5}.0`) with strict `git apply`, as a single invocation
- [x] Every touched file byte-compiles after patching
- [x] Hunk headers verified: declared line counts and offsets match content
- [x] No lines exceed the 88-column limit
- [x] Capture traces produced on GPU with the V2 runner active and no environment pin
- [x] Annotations parse under `CaptureAnnotation`

## Notes for review

- Indenting the capture body under the profiler pushed two pre-existing comments past 88 columns, so they are rewrapped. No logic in them changed.
- No file overlap with #755, which covers 0.14 through 0.21.
- 0.23 and 0.24 were validated statically but not on GPU; their `capture()` bodies are byte-identical to 0.25's, so the same patch text applies unchanged.

### DIFF
--- a/examples/custom_workflows/inference_analysis/vllm_patches/config_vllm_v0.22.0.patch
+++ b/examples/custom_workflows/inference_analysis/vllm_patches/config_vllm_v0.22.0.patch
@@ -31,6 +31,130 @@ index 68fa788..f0e29d0 100644
 +            )
 +
          return self
+diff --git a/vllm/v1/worker/gpu/cudagraph_utils.py b/vllm/v1/worker/gpu/cudagraph_utils.py
+index c7a7ffe..5a3b231 100644
+--- a/vllm/v1/worker/gpu/cudagraph_utils.py
++++ b/vllm/v1/worker/gpu/cudagraph_utils.py
+@@ -2,6 +2,7 @@
+ # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+ from collections import defaultdict
+ from collections.abc import Callable
++from contextlib import AbstractContextManager, nullcontext
+ from dataclasses import dataclass
+ from typing import Any, NamedTuple
+ 
+@@ -14,6 +15,7 @@ from vllm.config import VllmConfig
+ from vllm.config.compilation import CUDAGraphMode
+ from vllm.distributed.parallel_state import (
+     get_pp_group,
++    get_world_group,
+     graph_capture,
+     is_global_first_rank,
+ )
+@@ -199,6 +201,38 @@ class CudaGraphManager:
+             create_forward_fn: Factory that prepares inputs (OUTSIDE graph) and
+                 returns a tuple of (forward_fn, captured_attn_state).
+         """
++        # Setup torch profiler for graph capture traces (conditional)
++        profiler_config = self.vllm_config.profiler_config
++        local_rank = get_world_group().local_rank
++        profiler: AbstractContextManager[Any]
++        if local_rank == 0 and profiler_config.capture_torch_profiler:
++            trace_dir = profiler_config.torch_profiler_dir + "/capture_traces"
++            profiler = torch.profiler.profile(
++                activities=[
++                    torch.profiler.ProfilerActivity.CPU,
++                    torch.profiler.ProfilerActivity.CUDA,
++                ],
++                record_shapes=True,
++                profile_memory=True,
++                with_stack=True,
++                on_trace_ready=torch.profiler.tensorboard_trace_handler(
++                    trace_dir,
++                    worker_name=f"graph_capture_rank_{local_rank}",
++                    use_gzip=True,
++                ),
++            )
++            logger.info(
++                "Rank %d: Torch profiler enabled for CUDA graph capture, "
++                "traces will be saved to: %s",
++                local_rank,
++                trace_dir,
++            )
++        else:
++            profiler = nullcontext()
++            logger.info(
++                "Rank %d: Torch profiler disabled for CUDA graph capture", local_rank
++            )
++
+         captured_attn_states: dict[
+             BatchExecutionDescriptor, CapturedAttentionState
+         ] = {}
+@@ -224,32 +258,38 @@ class CudaGraphManager:
+                     logger.debug(
+                         "CG Capture: mode=%s, batch_desc=%s", desc.cg_mode.name, desc
+                     )
+-                    if desc.cg_mode == CUDAGraphMode.PIECEWISE:
+-                        captured_attn_states[desc] = attn_state
+-                        forward_fn(CUDAGraphMode.PIECEWISE)
+-                    else:
+-                        # Capture with fresh attention state. The warmup
+-                        # attention state is discarded because some backends
+-                        # (e.g. FlashMLA) perform lazy initializations that
+-                        # must be captured in the graph.
+-                        forward_fn, attn_state = create_forward_fn(desc)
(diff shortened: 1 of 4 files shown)


devalshahamd:
Thanks @tkarttunen  for the PR!

I am going through the patches, I think we need some additions to cover speculator and encoder models, and also annotations for that. I will try to build on top of this and merge both PRs after testing.

devalshahamd:
Hello @tkarttunen 

I had to make some changes to enable support for capturing encoder and speculators as well, and decided that having a common utility will be easier to get it upstreamed. I have merged #995 that includes v2 patches bases on that approach. I have tested it on a few models and versions on my side. Please let me know if those patches work for you. 

I have also created a PR to vLLM: https://github.com/vllm-project/vllm/pull/54061 using the same patches.

tkarttunen:
I know this is old, but I have deferred testing your merged #995 fix until we had time to rebase our code to pin to tracelens main and update our vllm patched. Looks good - your common-utility approach that includes capturing encoder and spectator is the right one!

Agentic code review detected, however, that tracelens main does not support the _consumption_ (analysis) of the captured traces yet. It claims that classify_graph_capture_trace cannot parse the V2 runner annotations yet, or that load_capture_folder cannot recognize the non-decoder shards. This is probably a known limitation at this moment, but I'm copy-pasting these findings nevertheless below.

AI GENERATED TEXT BELOW

## 1. `load_capture_folder` discards every non-decoder shard

`Trace2Tree/trace_capture_merge_experimental.py:613`:

```python
if not (mode in ["FULL", "PIECEWISE"] and isinstance(batch_size, int)):
    print("Warning: invalid batch size or mode, skipping: {}".format(entry))
    continue
```

`mode` comes from `CaptureAnnotation`, whose `CAPTURE_PATTERN` is `capture_(\d+)_(.*)` — so the
label ends up inside `mode`:

| annotation                    | `batch_size` | `mode`              | gate |
| ----------------------------- | ------------ | ------------------- | ---- |
| `capture_4_FULL`              | 4            | `'FULL'`            | PASS |
| `capture_4_draft_FULL`        | 4            | `'draft_FULL'`      | SKIP |
| `capture_512_encoder_default` | 512          | `'encoder_default'` | SKIP |

(measured by running `CaptureAnnotation` at `2526446f`, not read off the regex)

So the speculator and encoder shards your patches now produce are written, classified, and then
dropped with a warning before the merge. The decode ladder merges as before.

There's a trap in the obvious fix. If the gate is simply relaxed to accept the new modes, the key
becomes ambiguous rather than wrong-but-visible: the speculator captures **the same batch sizes as
the decoder**, and a few lines down the merge does

```python
result[key] = filepath
```

so `4_FULL` from the decoder and a normalised `4_FULL` from the speculator collide silently,
last-writer-wins. We hit exactly this in our own checker — it reported a false duplicate at `4_FULL`
— and fixed it by making the subsystem part of the identity rather than part of the mode string.
Something like parsing `capture_<number>[_<subsystem>]_<variant>` into three fields and keying on
`(subsystem, number, variant)` would work; the encoder in particular shouldn't share a key space
with the decode ladder at all, since its number is a token budget and not a batch size.

One small thing that made this harder to trace: for the speculator the filename suffix and the
annotation label disagree — the file is `graph_capture_rank_0_speculator...` while the annotation
says `draft`. Worth unifying if it's not load-bearing somewhere I haven't seen.

## 2. `classify_graph_capture_trace` can't parse annotations under the V2 runner

`Reporting/generate_perf_report_pytorch_inference.py:208`:

```python
dummy_run_pattern = re.compile(
    r"vllm/v1/worker/gpu_model_runner\.py\(\d+\): _dummy_run"
)
```

and the annotation path is taken only when

```python
if annotation_roots and len(annotation_roots) == len(dummy_roots):
```

Both runners define `_dummy_run` (`gpu_model_runner.py:5830` and `gpu/model_runner.py:555` at stock
`v0.27.1`), but the module path differs — under the V2 runner the frames read
`vllm/v1/worker/gpu/model_runner.py`, so `dummy_roots` is empty, the equality fails, and
classification falls through to:

```python
mode = infer_mode_from_captures(num_captures)   # "FULL" if num_captures <= 1 else "PIECEWISE"
batch_size = most_common_first_dim(events)
```

The net effect is that the annotations `#995` works to emit correctly are then not read on the very
code path `#995` enables. `execution_details.json` gets an inferred mode and an inferred batch size
instead. For the encoder that inference has no correct answer available, since there is no decode
batch size to find.

Adding the V2 module path as an alternation looks sufficient — note it has to be a real alternation
rather than an optional `gpu/`, since the legacy file is `gpu_model_runner.py` and not
`gpu/model_runner.py`:

```python
dummy_run_pattern = re.compile(
    r"vllm/v1/worker/(?:gpu_model_runner|gpu/model_runner)\.py\(\d+\): _dummy_run"
)
```

though you'd know better than me whether the `len(annotation_roots) == len(dummy_roots)` equality
still holds once the speculator and encoder contribute annotations that have no `_dummy_run` root of
their own — that's the part I'd expect to need care.

PR #956, opened 2026-08-20
Skip creating empty dfs

### LINKED ISSUE #482: For loop at line 325 in generate_perf_report_pytorch goes through dict_cat2names
There's a for loop that tries to create a dataframe for every operation category in dict_cat2names (which seems to be a pre-computed mapping), regardless of whether the operation categories are in the trace. The code doesn't fail, it just throws user warnings. But the user warnings could be confusing/alarming to the user. 
I feel like there should be a simple if condition in the for loop where if the operation category isn't in the trace it continues.

kyle-hoffmeyer:
Skip creating dataframes for empty data. This avoids warnings such as this:

```
TraceLens/TraceLens/TreePerf/tree_perf.py:557: UserWarning: Input list of events is empty. Returning an empty DataFrame.
  warnings.warn(
TraceLens/TraceLens/TreePerf/tree_perf.py:677: UserWarning: Input DataFrame is empty. Returning an empty summary DataFrame.
  warnings.warn(
```

Closes #482


tsrikris (approved):
LGTM conceptually!

PR #959, opened 2026-08-21
Fix comparative graph capture in evals

kyle-hoffmeyer:
Comparative graph capture support in evals was broken. Now capture folder paths are correctly extracted from trace csv and passed into prompt.

Verified that standalone and comparative prompts to agent are still correct.


agent_evals/Analysis/eval_scripts/run_repeatability_parallel.sh
```diff
@@ -245,10 +247,24 @@ run_single_job() {
...
     local capture_suffix=""
+    local analysis_mode="default"
+    if [[ -n "$capture_folder1" ]]; then
+        capture_suffix+=" capture folder for trace1 $capture_folder1"
+    fi
+    if [[ -n "$capture_folder2" ]]; then
+        capture_suffix+=" capture folder for trace2 $capture_folder2"
+    fi
+    if [[ -n "$capture_folder1" || -n "$capture_folder2" ]]; then
+        analysis_mode="inference"
```
Ahmedhasssan-aig, line 266: Why do we need to specifically mention the analysis_mode here?

agent_evals/Analysis/eval_scripts/run_repeatability_parallel.sh
```diff
@@ -503,7 +520,7 @@ build_combined_csv() {
         csv="$EVALS_DIR/analysis_tests/combined_traces_${scope}.csv"
         [[ -f "$csv" ]] || continue
         if [[ "$scope" == "comparative" ]]; then
-            while IFS=, read -r id sub t1 t2 ref plat plat2; do
+            while IFS=, read -r id sub t1 t2 ref plat plat2 cf1 cf2; do
```
Ahmedhasssan-aig, line 523: Where are cf1 and cf2 defined?

Ahmedhasssan-aig:
I left two comments for minor clarification, otherwise it LGTM

agent_evals/Analysis/eval_scripts/run_repeatability_parallel.sh
kyle-hoffmeyer: generate_perf_report_pytorch_inference needs to be called when the capture folders are being used. So we need to specify in the prompt that the analysis mode is inference

agent_evals/Analysis/eval_scripts/run_repeatability_parallel.sh
kyle-hoffmeyer: Just pushed. It doesn't need to be initialized since `read` initializes the variables and writes the variables, but I initialized the variables alongside the other variables for consistency.

Ahmedhasssan-aig (approved):
LGTM!

PR #967, opened 2026-08-25
patches for sglang 5.18 with the torchlib fix

devalshahamd:
Adding patches for SGLang-dev 5.18.0 image. This version has some of the TraceLens patches integrated (upstreamed PRs), so only shape profiling related patches are added. Another fix is to override torchlib so files to ensure the rocm 7.2.4 so files are used to enable profiling of graph captured region.


mohbasit (approved):
Tests with DeepSeek also passed, looks good

PR #969, opened 2026-08-26
Fix formatting issues in performance report documentation

mattwill-amd:


amitkumar-amd (approved):

ajassani:
Thanks for the style pass and for moving the notebooks to `notebooks/` rather than dropping them. The RTD preview looks good. A few things before merge:

**Blocking**

1. Links around the notebook move are stale, and that is the path users will actually click now that these live on GitHub rather than the docs site.

   Inbound still point at `docs/tutorials/`, which this PR removes:
   - `README.md` (Quick Start and Additional Resources) → `docs/tutorials/torch-profiling.ipynb`
   - `TraceLens/Agent/Analysis/README.md` → `../../../docs/tutorials/torch-profiling.ipynb`

   Please retarget those to `notebooks/torch-profiling.ipynb`. The trace-capture note removed from `docs/how-to/generate-perf-report-pytorch.md` should also point at the new location (a GitHub `blob` URL is fine now that it is outside the docs tree).

   The notebooks themselves also have broken relative links:
   - `notebooks/torch-profiling.ipynb` → `./profiling_analysis.md`
   - `notebooks/anomaly-detection.ipynb` → `./torch_profiling_guide.ipynb`, `../generate_perf_report.md`, `./torch_profiling_analysis.md`

   `./torch_profiling_guide.ipynb` should be `./torch-profiling.ipynb`. The `.md` targets should be GitHub URLs to the conceptual/how-to pages (they are not siblings of `notebooks/`).

2. `docs/how-to/agent.md` changes the meaning of the analysis-mode guidance. Original `non-vLLM/SGLang/ATOM` means eager inference that is *not* those three. The slash-to-comma rewrite:

   > `default` for training and non-vLLM, SGLang, or ATOM eager inference

   reads as though SGLang and ATOM eager inference use `default`, which contradicts the next clause. Suggested wording: "`default` for training and eager inference outside vLLM, SGLang, and ATOM".

**Non-blocking**

- `docs/how-to/sdk-analysis.md`: the example DataFrame headers were restyled to `Type` / `Time (ms)` / `Percent`. The API still returns `type`, `time ms`, and `percent` (`get_df_gpu_timeline`). Prefer the actual column names so the table matches what the user sees.
- Please restore the `Related topics` section removed from `docs/conceptual/triton-perf-model-walkthrough.md`.

The rest of the formatting and figure-directive updates look good to me.

gabeweisz (approved):

ajassani (approved):

PR #970, opened 2026-08-26
TL Agent Deterministic Fallback

tsrikris:
# Deterministic (no-LLM) fallback for graph-collapsed traces

Some inference traces are **graph-under-recorded**: a GPU graph replay or a
compiled region collapses the whole workload behind a single launch, so the
profiler records device kernels but not the per-op decomposition (Python / aten
op, shapes, launcher path) the analysis pipeline needs. On these traces per-op
decomposition fails and the LLM analysis path cannot run.

This change adds a deterministic fallback that still emits a parser-compatible
`analysis.md`. It recovers the only three things a graph-collapsed trace carries
(**kernel name, time, and %E2E**) and marks every other field as unrecoverable
instead of guessing at it.

## Architecture

```
                    Step 1: perf report ──> perf_report_csvs/
                                                   │
                                                   ▼
                    Step 2: check_graph_replay_coverage()  ── graph_replay_fraction
                                                   │
                        ┌──────────────────────────┴──────────────────────────┐
              frac ≤ MAX │ (healthy)                          frac > MAX │ (graph-collapsed)
                         ▼                                                ▼
              Steps 3-6: prepare category data          render_fallback_report()
              Steps 7-8: analysis sub-agents                        │
              Step 12: analysis.md (full)          ┌────────────────┴──────────────┐
                         │                          │ _load_surviving_rows          │
                         │                          │  (drop plumbing wrappers)     │
                         │                          │ rank device kernels by %E2E   │
                         │                          │ apply %E2E floor + count cap  │
                         │                          │ unrecoverable cells = "—"     │
                         │                          │ Kernel Name  = raw symbol     │
                         │                          │ P-item heading = normalized   │
                         │                          └────────────────┬──────────────┘
                         ▼                                           ▼
                 analysis.md (full)                       analysis.md (degraded)
                         └────────────────────┬──────────────────────┘
                                              ▼
                        same contract: #### P{rank}: / reasoning-candidate /
                        impact-begin kind=p_item / 9-column **Data:** table
                                              │
                                              ▼
                               downstream analysis.md consumer
```

The fallback branch is a drop-in: it emits the **same** `analysis.md` marker and
table contract as the full analysis path, so a downstream consumer that parses
`analysis.md` needs no structural change.

## Design

- **Trace-quality gate (Step 2).** After perf-report generation,
  `check_graph_replay_coverage()` reads the already-written `perf_report_csvs/`,
  computes the fraction of device time hidden behind graph-replay / compiled-region
  launches, and trips when it exceeds `GRAPH_REPLAY_FRACTION_MAX`. Benign
  eager-launch and memcpy plumbing wrappers (present in healthy traces too) are
  excluded, so the signal isolates the pathology.
- **Deterministic report writer.** On a bad verdict, `render_fallback_report()`
  builds P-items from the surviving device-kernel rows, ranks them by %E2E, and
  writes the contract the pipeline already emits: `#### P{rank}:` headings, one
  `reasoning-candidate` marker and one `impact-begin kind=p_item` marker per P-item,
  and the 9-column `**Data:**` table. No LLM, no GPU, fully deterministic.
- **Honest unrecoverable cells.** Fields that a graph-collapsed trace never captured
  (Operation, Args, Kernel Path, Count, FLOPS/Byte, Efficiency, Bound) render a
  literal `—` instead of a fabricated or duplicated value. The raw device symbol is
  preserved verbatim in the `Kernel Name` cell so a downstream consumer can key on it,
  and the P-item heading shows a display-shortened form of that symbol.
- **P-item count control.** A `%E2E` floor (`MIN_PITEM_PERCENT_E2E`) plus a defensive
  count cap (`MAX_PITEM_COUNT`) collapse noisy traces to the significant few. The
  percentage denominator is built over **all** surviving rows before filtering, so
  dropping the tail never inflates the survivors. Drops are reported in the banner
  (never silent) when non-zero.
- **Display-only name normalization.** `_normalize_kernel_name()` shortens the P-item
  heading to a fixed length with a truncation marker. It never demangles and never
  strips qualifiers, so the symbol in `Kernel Name` and the impact-grouping key both
  stay raw. Two long GEMM variants that differ only past the truncation boundary
  collapse to the same heading but remain distinct P-items in the arithmetic.
- **Degraded banner.** A visible banner states the graph-replay fraction, what is
  recoverable (kernel, time, %E2E), and what was never captured (shapes, launcher
  path, quant operands, category, efficiency), so the report can never be mistaken for
  a full analysis.
- **CSV field-size hardening.** The gate reads every trace's CSVs, and traces with
  very large cells exceeded `csv`'s default field limit and aborted the gate before it
  could pass. A module-level `csv.field_size_limit(...)` is now raised once with a
  platform-safe clamp (start high, halve on `OverflowError`).
- **Pipeline step renumber.** Inserting the gate as Step 2 shifted the rest of the
  pipeline by one, so the orchestrator prep, the validation utility, and the
  orchestrator skill / reference / template prose were renumbered in lockstep. This is
  prose and print-label only, with no logic change.

## Validation

- **Unit tests (CI):** `tests/test_analysis_agent_deterministic.py` covers the gate
  (fraction math, plumbing-wrapper exclusion, field-size case), the writer contract
  (markers 1:1 with P-items, unrecoverable cells em-dashed, raw kernel name preserved,
  normalized heading), the normalizer matrix (identity / truncate / mangled-stays-raw
  / length boundary / never-empty), the floor/cap boundaries and drop reporting, and
  the consumer-side substitution of the kernel name when Operation is an em dash. No
  GPU, no LLM.
- **End-to-end:** exercised through the full orchestrator on several graph-collapsed
  inference traces spanning distinct workload shapes (dense and MoE decode, diffusion
  image generation, speculative decode). Each tripped the Step 2 gate and routed to
  the fallback, and every report preserved the marker contract
  (`reasoning-candidate` = `impact-begin` = data-tables = P-item count) with the floor
  and cap applied.
- **Coverage vs the full analysis path:** on traces where a non-fallback baseline
  exists, the fallback recovers the complete compute-heavy tail (every dominant GEMM,
  attention, quantization, sort, and norm kernel), and surfaces additional
  kernel-granular items the category-grouped path folds together. The only kernels it
  cannot recover are ones that live inside a fused region and are visible only when
  capture-enabled decomposition splits them apart. That is a genuine limit of a
  graph-collapsed trace, and the `—` cells reflect it honestly.
- **Healthy no-op:** on healthy traces the gate returns false and the writer never
  fires.

## Downstream / contract impact

The change is additive and parser-compatible. The 9-column table header, the
`#### P{rank}:` / `reasoning-candidate` / `impact-begin` markers, and their attribute
names are all unchanged. The one behavioral note for a downstream consumer is the `—`
`Operation` cell: a consumer that today drops rows with an empty Operation should
substitute the `Kernel Name` symbol for those rows (a backward-compatible reader
relaxation) so that fallback reports parse to the intended candidates.

## Test plan

- [x] `pytest tests/test_analysis_agent_deterministic.py`
- [x] `black --check` on all changed Python (clean)
- [x] Copyright headers valid on all changed files
- [x] End-to-end fallback on multiple graph-collapsed inference traces
- [x] Coverage compared against the full analysis path
- [x] Field-size fix verified on a large-cell trace
- [x] Healthy no-op returns false at the gate

### DIFF
--- a/TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md
+++ b/TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md
@@ -263,14 +263,12 @@ All commands below append `<suffix_1>` and `<suffix_2>`, resolved by `<compariso
 
 ```bash
 <prefix> python3 -c \"
-import sys
 from TraceLens.Agent.Analysis.utils import check_graph_replay_coverage
-v = check_graph_replay_coverage(sys.argv[1])
-status = 'GRAPH_UNDER_RECORDED' if v.graph_under_recorded else 'OK'
+coverage = check_graph_replay_coverage(<unified_perf_csv>)
+status = 'GRAPH_UNDER_RECORDED' if coverage.graph_under_recorded else 'OK'
 print(f'STATUS={status}')
-print(f'GRAPH_REPLAY_FRACTION={v.graph_replay_fraction}')
-print(f'REASON={v.reason}')
-\" '<unified_perf_csv>'
+print(f'GRAPH_REPLAY_FRACTION={coverage.graph_replay_fraction}')
+" 
 ```
 <graph_replay_fraction> refers to the returned GRAPH_REPLAY_FRACTION value.
 
@@ -282,7 +280,7 @@ print(f'REASON={v.reason}')
    <prefix> python3 -m TraceLens.Agent.Analysis.utils.deterministic_fallback \
      --unified-perf-csv '<unified_perf_csv>' \
      --output-dir '<output_dir>' \ 
-     --graph-replay-fraction '<graph_replay_fraction>'
+     --graph-replay-fraction <graph_replay_fraction>
    ```
 3. **Skip Steps 3-12.** The writer already produced `<output_dir>/analysis.md`
 
--- a/TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md
+++ b/TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md
@@ -10,7 +10,7 @@ This document is the detailed specification for the TraceLens **analysis-orchest
 
 ## Workflow overview
 
-The orchestrator runs a staged pipeline (Steps 0–11): collect inputs and environment prefix, generate perf reports, prepare category data via `orchestrator_prepare.py`, run system-level and compute-kernel subagents in parallel, aggregate and validate findings, identify the model, render plots when no extension is present, and write `analysis.md` via remote `tee` heredocs. Only Steps 6, 7, and 9 delegate to Task subagents; all other steps run in the main agent.
+The orchestrator runs a staged pipeline (Steps 0–12): collect inputs and environment prefix, generate perf reports, prepare category data via `orchestrator_prepare.py`, run system-level and compute-kernel subagents in parallel, aggregate and validate findings, identify the model, render plots when no extension is present, and write `analysis.md` via remote `tee` heredocs. Only Steps 7, 8, and 10 delegate to Task subagents; all other steps run in the main agent.
 
 ---
 
@@ -27,18 +27,19 @@ Use vendor-agnostic terminology throughout such as GPU kernels, collective commu
 ```
 0. Query User Inputs (Platform, Trace Path(s), Analysis Mode, Environment Setup)
 1. Generate Performance Report (branches on analysis mode: training vs inference then, comparison scope)
-2-5. Prepare Category Data (GPU Util, Top Ops, Tree Data, Multi-Kernel Data, Category Filtering)
-6. System-Level Analysis (PARALLEL, CPU/Idle + Kernel Fusion + Multi-Kernel) → system_findings/
-7. Invoke Compute Kernel Subagents (PARALLEL, read category_findings[] from _metrics.json) → category_findings/
-   7.5. Aggregate per-category category_findings[] → priority_data.json::findings[] (globally sorted)
-8. Validate Subagent Outputs (system_findings/ + category_findings/)
-9. Prepare Report Data (load_findings) + Model Identification (subagent) → metadata/model_info.json
-10. Render performance PNG IF agent_extension.py is absent
-11. Generate Final Report (composable System + Compute sections), validate it,
+2. Trace-Quality Gate (deterministic fallback branch for graph-under-recorded traces)
+3-6. Prepare Category Data (GPU Util, Top Ops, Tree Data, Multi-Kernel Data, Category Filtering)
+7. System-Level Analysis (PARALLEL, CPU/Idle + Kernel Fusion + Multi-Kernel) → system_findings/
+8. Invoke Compute Kernel Subagents (PARALLEL, read category_findings[] from _metrics.json) → category_findings/
+   8.5. Aggregate per-category category_findings[] → priority_data.json::findings[] (globally sorted)
+9. Validate Subagent Outputs (system_findings/ + category_findings/)
+10. Prepare Report Data (load_findings) + Model Identification (subagent) → metadata/model_info.json
+11. Render performance PNG IF agent_extension.py is absent
+12. Generate Final Report (composable System + Compute sections), validate it,
     optionally invoke agent_extension.py (when present), then embed the PNG into the report.
 ```
 
-**Subagent usage:** Only invoke Task subagents in steps that explicitly say "subagent" (Steps 6, 7, 9). All other steps (including Step 7.5) must be performed directly by the orchestrator using the command prefix.
+**Subagent usage:** Only invoke Task subagents in steps that explicitly say "subagent" (Steps 7, 8, 10). All other steps (including Step 8.5) must be performed directly by the orchestrator using the command prefix.
 
 ---
 
@@ -255,7 +256,35 @@ All commands below append `<suffix_1>` and `<suffix_2>`, resolved by `<compariso
 
 ---
 
-## Steps 2-5: Prepare Category Data
+## Step 2: Trace-Quality Gate
+
+
+`<csv_dir>` is `<output_dir>/perf_report_csvs` (`standalone`) or `<output_dir>/perf_report_trace1_csvs` (`comparative`, primary trace).
+
+```bash
+<prefix> python3 -c \"
+import sys
+from TraceLens.Agent.Analysis.utils import check_graph_replay_coverage
+v = check_graph_replay_coverage(sys.argv[1])
+print('GRAPH_UNDER_RECORDED' if v.graph_under_recorded else 'OK', v.graph_replay_fraction, v.reason)
+\" '<csv_dir>'
+```
+
+**`OK`:** Proceed
+**`GRAPH_UNDER_RECORDED`:**
+1. Emit `[DIAG:trace_quality:GRAPH_UNDER_RECORDED]`: Deterministic fallback report
+2. Run the deterministic fallback writer on `<csv_dir>`:
+   ```bash
+   <prefix> python3 -m TraceLens.Agent.Analysis.utils.deterministic_fallback \
+     --perf-csvs-dir '<csv_dir>' \
+     --output-dir '<output_dir>' \
+     --graph-replay-fraction <graph_replay_fraction>
+   ```
+3. **Skip Steps 3-12.** The writer already produced `<output_dir>/analysis.md`
+
+---
+
+## Steps 3-6: Prepare Category Data
 
 Execute the TraceLens Agentic Mode orchestrator preparation script:
 
@@ -269,11 +298,11 @@ Execute the TraceLens Agentic Mode orchestrator preparation script:
 ```
 
 This script performs:
-- **Step 2:** Assess GPU utilization (computation, idle, communication times)
-- **Step 3:** Identify top 10 operations by GPU time
-- **Step 4:** Pre-compute tree data for bottleneck operations (load trace ONCE)
-- **Step 4.5:** Pre-compute multi-kernel issue data (memcpy by direction, NCCL events, overlap metrics)
-- **Step 5:** Filter and export category-specific data
+- **Step 3:** Assess GPU utilization (computation, idle, communication times)
+- **Step 4:** Identify top 10 operations by GPU time
+- **Step 5:** Pre-compute tree data for bottleneck operations (load trace ONCE)
+- **Step 5.5:** Pre-compute multi-kernel issue data (memcpy by direction, NCCL events, overlap metrics)
+- **Step 6:** Filter and export category-specific data
 
 **Outputs:**
 - `category_data/<category>_ops.csv` - Filtered operations per category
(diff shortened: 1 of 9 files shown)


tests/test_analysis_agent_deterministic.py
```diff
@@ -0,0 +1,688 @@
...
+    "bound",
+)
+
+_EM_DASH = "—"
+
+
+# ---------------------------------------------------------------------------
+# Fixture helpers
+# ---------------------------------------------------------------------------
+def _make_perf_csv(tmp_path, rows):
```
gabeweisz, line 69: I feel like this function should not be necessary. But I guess this is a test so maybe not a big deal

tests/test_analysis_agent_deterministic.py
```diff
@@ -0,0 +1,688 @@
...
+    ``rows`` is a list of ``(name, weight, percent)`` tuples. Only the three
+    columns the code reads are written; ``percent`` defaults to 0.0 if omitted.
+    """
+    csv_path = tmp_path / "unified_perf_summary.csv"
+    with open(csv_path, "w", newline="", encoding="utf-8") as fh:
+        writer = csv.writer(fh)
+        writer.writerow(["name", _WEIGHT_COLUMN, _PERCENT_COLUMN])
+        for row in rows:
+            name, weight = row[0], row[1]
+            percent = row[2] if len(row) > 2 else 0.0
```
gabeweisz, line 83: It looks like the code that generates the rows has to generate 3 inputs - so isn't missing the third columns an error?

tests/test_analysis_agent_deterministic.py
tsrikris: I think we can include common test utils since we end up having to create CSVs to get to the 95% coverage bar. Let me know if this is gating this PR

kyle-hoffmeyer:
@tsrikris could you include an example of a deterministic report?

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+    _KERNEL_NAME_TRUNC_LEN,
+)
+
+# Real traces have CSV cells far larger than csv's 131072-char default, which
+# would crash the trace-quality gate; lift the limit to the platform maximum.
+csv.field_size_limit(sys.maxsize)
+
+# Constants
+
+GRAPH_REPLAY_FRACTION_MAX = 0.10
```
kyle-hoffmeyer, line 46: Do we want the fallback to run when just 10% of total time is graph replay? Is this not too low? What do you think

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+_GRAPH_REPLAY_WRAPPER_PREFIXES = ("hipGraphLaunch", "cudaGraphLaunch")
+_TORCH_COMPILED_PREFIX = "Torch-Compiled Region"
+
+# Generic launch/memcpy shims that are Synthetic Ops but benign runtime plumbing.
+# Excluded from both the graph-replay signal and the
+# fallback candidate list.
+_PLUMBING_WRAPPER_PREFIXES = frozenset(
+    {"hipLaunchKernel", "hipModuleLaunchKernel", "hipMemcpyAsync"}
+)
+_PLUMBING_WRAPPER_STARTSWITH = ("Memcpy", "__amd_rocclr_")
```
kyle-hoffmeyer, line 68: We should add cuda prefixes for these two sets

TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md
```diff
@@ -255,7 +256,35 @@ All commands below append `<suffix_1>` and `<suffix_2>`, resolved by `<compariso
...
+
+**`OK`:** Proceed
+**`GRAPH_UNDER_RECORDED`:**
+1. Emit `[DIAG:trace_quality:GRAPH_UNDER_RECORDED]`: Deterministic fallback report
+2. Run the deterministic fallback writer on `<unified_perf_csv>`:
+   ```bash
+   <prefix> python3 -m TraceLens.Agent.Analysis.utils.deterministic_fallback \
+     --unified-perf-csv '<unified_perf_csv>' \
+     --output-dir '<output_dir>' \
+     --graph-replay-fraction <graph_replay_fraction>
```
kyle-hoffmeyer: <graph_replay_fraction> is never explicitly defined and agent needs to infer it. Just want to make sure in your testing the agent has been able to infer this definition.

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--unified-perf-csv", type=Path, required=True)
+    parser.add_argument("--output-dir", type=Path, required=True)
+    parser.add_argument("--graph-replay-fraction", type=float, default=None)
+    args = parser.parse_args()
+
+    frac = args.graph_replay_fraction
+    if frac is None:
+        frac = check_graph_replay_coverage(args.unified_perf_csv).graph_replay_fraction
```
kyle-hoffmeyer, line 250: If we don't already know graph replay coverage we check for it here. But it seems to render the fallback report regardless of what the graph replay coverage is. Currently the deterministic fallback is only called from TraceLens Agent when it's graph-under-recorded so it's fine, but just want to point out that this will create a fallback report regardless of what graph-replay-fraction is.

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+def _is_graph_replay_wrapper(name: str) -> bool:
+    if "->" not in name:
+        return False
+    prefix = name.split("->", 1)[0]
+    return prefix in _GRAPH_REPLAY_WRAPPER_PREFIXES or prefix.startswith(
+        _TORCH_COMPILED_PREFIX
+    )
+
+
+def _surviving_rows(csv_rows):
```
kyle-hoffmeyer, line 191: _surviving_rows seems to assume every row's "name" column is a synthetic op / kernel which is true if the entire trace is graph-replay, but for non synthetic ops, the "name" column will refer to the CPU op. Functionally, it's fine since "name" is always what appears in the P-items regardless of its a synthetic kernel op or a cpu op, but just want to highlight this since the variable "kernel" may be little misleading here

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+        lines.append("")
+        lines.append("**Data:**")
+        lines.append("")
+        lines.append(_TABLE_HEADER)
+        lines.append(_TABLE_DIVIDER)
+        time_ms = row["weight"] / 1000.0
+        lines.append(
+            f"| — | — | — | {kernel} | {time_ms:.3f} | "
+            f"{row['percent']:.2f} | — | — | — | "
```
kyle-hoffmeyer: Is there a reason we don't include count here? Isn't count derivable?

kyle-hoffmeyer (commented):
Looks good. Minor comments

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
tsrikris: Right now, this fallback is bare-bones; handles a binary scenario where things are all right or really bad. The intermediate scenario still could exist. Handling this hybrid path (Agent for locations where the trace is linked with graph capture and deterministic for the others and fusing them is tricky). I'd say hopefully the trace splitter is able to ensure that we always fall into one or the other. At the end of the day this PR enables Hyperloom to not break completely while we work on fixing the graph capture issues.

TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md
tsrikris: Good flag! I've removed it completely since it seems to be redundant.

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
tsrikris: I'd say this is linked to the 10% graph threshold. I'm hoping that we get another opportunity to improve this once more mixed traces come in

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
tsrikris: Done!

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
tsrikris: I'm hoping the user calls this feauture only when necessary (i.e via the TraceLens Agent)

tests/test_analysis_agent_deterministic.py
tsrikris: Updated the test!

gabeweisz (approved):

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+    rows = []
+    for row in csv_rows:
+        name = row.get("name", "")
+        if "->" in name:
+            prefix, tail = name.split("->", 1)
+            if _is_plumbing_wrapper(prefix):
+                continue
+            kernel = _strip_synthetic_suffix(tail)
+        else:
+            kernel = _strip_synthetic_suffix(name)
```
devalshahamd, line 192: The unified perf report also has kernel names. Can we use those?

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
```diff
@@ -0,0 +1,247 @@
...
+            f"low={low} mid={mid} high={high} -->"
+        )
+        lines.append("")
+        lines.append("**Data:**")
+        lines.append("")
+        lines.append(_TABLE_HEADER)
+        lines.append(_TABLE_DIVIDER)
+        time_ms = row["weight"] / 1000.0
+        lines.append(
+            f"| — | — | — | {kernel} | {time_ms:.3f} | "
```
devalshahamd, line 164: We might want to add kernel as Operation or change the kernel_candidates.json file generation logic since it skips a row of Operation name is empty.

devalshahamd (commented):
The logic looks good. We need to test some examples where we have a mix of graph captured and eager regions and test which route works better for that. We can follow up on that in the next PR.

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
tsrikris: Noted for a future update!

TraceLens/Agent/Analysis/utils/deterministic_fallback.py
tsrikris: So this is from the unified per report. The name column just has kernel names prefixed with something like hipgraphlaunch -> <kernel_name>

PR #977, opened 2026-08-28
Use full git+https path for pip install in reference.md

### LINKED ISSUE #976: `analysis-orchestrator` points to unrelated TraceLens PyPi Package
# Description

I was poking at the tracelens skill and noticed something in `reference.md` Step 0. 
https://github.com/AMD-AGI/TraceLens/blob/main/TraceLens/Agent/Analysis/skills/analysis-orchestrator/reference.md

It tells the agent to run `pip install TraceLens` if the import gate fails, but that PyPI name belongs to an unrelated project (github.com/ssf0409/tracelens, an agent-eval framework).
https://pypi.org/project/tracelens/

It installs fine, then import TraceLens still fails. And if AMD's TraceLens is already installed, pip just says "already satisfied" and does nothing, since upstream's setup.py declares the same name.

## Reproducing
Create a fresh venv + pip install TraceLens + python -c "import TraceLens"

gabeweisz:
Fixes #976


gabeweisz:
@tsrikris says he is OK with me bypassing rules to merge this. Tests have passed

PR #979, opened 2026-08-28
feat(kernel-source-mapping): Add GPU kernel source mapping using active finder approach

Ahmedhasssan-aig:
A self-contained, Python package (`TraceLens/TraceUtils/kernel_source/`) that takes a GPU kernel name from a trace and returns the editable source file and line where it's defined, plus whether it's patchable. It looks kernels up by symbol rather than by file path, so it keeps working even when frameworks are reinstalled or upgraded. The package is fully additive and nothing in TraceLens imports it yet, so it can't break anything.

### New package — `TraceLens/TraceUtils/kernel_source/` (11 files)
| File | Purpose |
|---|---|
| `__init__.py` | Public API exports |
| `datatypes.py` | Shared dataclasses (dep-free, breaks import cycles) |
| `demangle.py` | 3-tier symbol demangling |
| `patchability.py` | Level-0 gate (no I/O) |
| `editable.py` | Editable-vs-generated path classifier (includes vLLM `torch_compile_cache`/`inductor_cache` marker fix) |
| `index.py` | csrc symbol indexer + cache (includes comment-stripping fix) |
| `resolver.py` | Level-1 native symbol resolution |
| `triton_pin.py` | Triton `.py` AST line-pinning |
| `discovery.py` | Active-finder framework tree discovery |
| `contract.py` | Public contract surface |
| `cli.py` | `TraceLens_resolve_kernel_source` entry point |

### Tests — `tests/` (6 files)
- `test_kernel_source_pipeline.py` — 58 tests, end-to-end across all 9 stages (comprehensive suite)
- `test_kernel_source_contract.py`, `_discovery.py`, `_gate.py`, `_index.py`, `_triton.py` — focused per-module tests

### Tooling (1 file)
- `tools/validate_kernel_source.py` — real-world parity harness (verdicts vs ground-truth `patchable`)

### Modified (1 file)
- `setup.py` — adds optional `kernel_source` extra (`itanium-demangler`) + `TraceLens_resolve_kernel_source` console script

## How downstream tooling calls this for source mapping
Downstream optimization tooling consumes this as a library via a thin adapter, selected behind an opt-in env flag:

```python
from TraceLens.TraceUtils.kernel_source import resolve
result = resolve(
    "<KERNEL_NAME>",                 # kernel_name: device symbol from the trace (mangled or plain)
    ["<SEARCH_PATH_1>", "..."],      # search_paths: dirs to search; pass None to auto-discover
    op_name="<OP_NAME>",             # optional: launching op name, helps the gate
    call_stack=[],                   # optional: call-stack frames, helps the gate
    run_gate=True,                   # optional: set False to skip the gate and resolve directly
)
print(result.source_file, result.line, result.patchable)
```

Flow: gate → resolve survivors against the currently-installed framework trees → normalize the verdict into the caller's fields. Validated on a real 35-kernel trace: 7 hit / 7 CK-gated / 19 pass-through / 0 disagreements vs the legacy resolver.

## Running the tests
```bash
pip install -e '.[dev,kernel_source]'
python -m pytest tests/test_kernel_source_pipeline.py -v     # 58 tests, self-contained
```

Minimal container (no heavy deps required):
```bash
docker run --rm -v "$PWD":/work -w /work python:3.11-slim bash -lc "pip install -q pytest && python -m pytest tests/test_kernel_source_pipeline.py -v"
```


TraceLens/TraceUtils/kernel_source/patchability.py
tsrikris, line 1: Should this stay within Hyperloom? That way if anyone updates kernel agents it's automatically tracked

TraceLens/TraceUtils/kernel_source/index.py
```diff
@@ -0,0 +1,388 @@
...
+]
+
+# Native source extensions to scan. Kept in sync with the editability filter in
+# :mod:`.editable`: a ``__global__`` def indexed from an extension the resolver
+# would later reject as non-editable is dead weight, so both lists must agree.
+_NATIVE_EXTS = (".cu", ".cuh", ".hip", ".h", ".hpp")
+
+# --- kernel-definition scanning ---------------------------------------------
+# A definition head is ``__global__`` <attrs / return type> NAME ( params ) { .
+# The tricky part is attributes that carry their own parentheses -- notably
```
tsrikris: Suggest trimming comment

TraceLens/TraceUtils/kernel_source/index.py
```diff
@@ -0,0 +1,388 @@
...
+            json.dump(asdict(index), fh)
+        tmp.replace(path)
+    except OSError as exc:
+        # Best-effort cache write; a failure here must not break index build/use.
+        log.debug("kernel index: cache write failed (%s): %s", index.fingerprint, exc)
+
+
+# Per-fingerprint in-process singletons, so repeated resolves for the same search
+# paths within one run do not re-read the on-disk cache.
+_PROCESS_INDEX: dict[str, SourceIndex] = {}
```
tsrikris, line 344: Suggest aggregating all env vars up top

TraceLens/TraceUtils/kernel_source/index.py
```diff
@@ -0,0 +1,388 @@
...
+        fingerprint=fingerprint(paths),
+        symbol_index=symbol_index,
+        build_ms=round((time.perf_counter() - started) * 1000.0, 2),
+        file_count=file_count,
+        symbol_count=len(symbol_index),
+    )
+
+
+# --- cache ------------------------------------------------------------------
+def _cache_path(fp: str) -> Path:
```
tsrikris, line 295: Should these functions below belong to a single class? What do you think?

TraceLens/TraceUtils/kernel_source/index.py
```diff
@@ -0,0 +1,388 @@
...
+    """Cache file path (dir from ``$TRACELENS_KSI_CACHE_DIR`` or a temp subdir).
+
+    When falling back to the system temp root (typically a shared ``/tmp`` on a
+    multi-user host), the subdir is scoped to the current user and created
+    owner-only (0o700) so users cannot collide on each other's cache. An explicit
+    ``$TRACELENS_KSI_CACHE_DIR`` is used verbatim.
+    """
+    raw = os.environ.get("TRACELENS_KSI_CACHE_DIR", "").strip()
+    if raw:
```
tsrikris, line 298: Can this segment of code be simplified? A simpler fallback?

TraceLens/TraceUtils/kernel_source/editable.py
```diff
@@ -0,0 +1,60 @@
...
+This is a pure path-shape check (no filesystem access), reused by both the
+native resolver and the Triton resolver.
+"""
+
+from __future__ import annotations
+
+__all__ = ["is_editable_source"]
+
+# Native device-code extensions that are always editable.
+_NATIVE_SOURCE_EXTS = (".cu", ".cuh", ".hip", ".h", ".hpp")
```
tsrikris, line 26: Can this be set as an arg with default values? In view of long term utilization where people start adding more editable sources

TraceLens/TraceUtils/kernel_source/discovery.py
```diff
@@ -0,0 +1,247 @@
...
+# ----------------------------------------------------------------------------
+def _version(name: str, root: Path | None) -> str:
+    """Installed version: dist metadata first, then a ``_version.py`` literal.
+
+    The source fallback covers packages that ship no dist metadata but record a
+    version file; it avoids importing the package (which can require a GPU at
+    import time).
+    """
+    try:
+        return importlib_metadata.version(name)
```
tsrikris: Imports up top as well

TraceLens/TraceUtils/kernel_source/discovery.py
```diff
@@ -0,0 +1,247 @@
...
+import re
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+
+__all__ = ["FrameworkRoot", "discover_frameworks", "discover_library_paths"]
+
+# Serving frameworks always located by name (for version reporting), even when
+# they ship no native source.
+_KNOWN = ("vllm", "sglang", "aiter")
```
tsrikris: Is ATOM also expected to be added here?

TraceLens/TraceUtils/kernel_source/discovery.py
```diff
@@ -0,0 +1,247 @@
...
+    return out
+
+
+def _discover_only() -> set[str]:
+    """Parse ``$TRACELENS_DISCOVER_ONLY`` into a lowercase allowlist (may be empty)."""
+    raw = os.environ.get("TRACELENS_DISCOVER_ONLY", "")
+    return {n.strip().lower() for n in raw.split(",") if n.strip()}
+
+
+def _spec_root(name: str) -> Path | None:
```
tsrikris: Another confirm if these independent functions need to be grouped into a class

tsrikris (commented):
Initial feedback!

TraceLens/TraceUtils/kernel_source/patchability.py
Ahmedhasssan-aig: If we are making this generic utility for Hyperloom, Magpie etc. I think we should keep it as part of Tracelens. This check is a gate before starting the source search

TraceLens/TraceUtils/kernel_source/index.py
Ahmedhasssan-aig: Done

TraceLens/TraceUtils/kernel_source/index.py
Ahmedhasssan-aig: Using these functions explicitly will avoid self/instantiation, so I would prefer to keep them as functions instead of making them part of a single class

TraceLens/TraceUtils/kernel_source/index.py
Ahmedhasssan-aig: Done

setup.py
```diff
@@ -65,6 +65,11 @@ def _wheel_version():
         "comparative": [
             "slodels[openai,anthropic,google-genai]",
         ],
+        # Optional: robust pure-Python demangling for the kernel-source resolver.
```
tsrikris: Suggest removing text

TraceLens/TraceUtils/kernel_source/cli.py
```diff
@@ -0,0 +1,137 @@
...
+        metavar="DIR",
+        help="A directory to search for native sources (repeatable). "
+        "When omitted, defaults are auto-discovered.",
+    )
+    parser.add_argument(
+        "--op-name",
+        default="",
+        help="Launching op name (used by the gate, e.g. MIOpen).",
+    )
+    parser.add_argument(
```
tsrikris: Could you elaborate what the role of this line is? Instead can we not pass in the structure in the TraceLens perf report column directly as a list?

TraceLens/TraceUtils/kernel_source/patchability.py
tsrikris: Sounds good, follow up q. What do you think might be the strategy to allow more editable sources in the future?

TraceLens/TraceUtils/kernel_source/index.py
```diff
@@ -0,0 +1,576 @@
...
+        if text[i] == "(":
+            depth += 1
+        elif text[i] == ")":
+            depth -= 1
+            if depth == 0:
+                return i + 1
+    return len(text)
+
+
+def _strip_comments(text: str) -> str:
```
tsrikris, line 130: Is something Regex based insufficient here?

 _COMMENT_RE = re.compile(
      r'//[^\n]*'          # line comment
      r'|/\*.*?\*/'        # block comment (non-greedy)
      r'|"(?:\\.|[^"\\])*"' # string literal
      r"|'(?:\\.|[^'\\])*'",# char literal
      re.DOTALL,
  )

TraceLens/TraceUtils/kernel_source/index.py
tsrikris: Can you check again? PROCESS_INDEX seems to be here still

TraceLens/TraceUtils/kernel_source/demangle.py
```diff
@@ -0,0 +1,211 @@
...
+    n = _rstrip_balanced(n, "(", ")")
+    n = _strip_trailing_qualifiers(n)
+    n = _rstrip_balanced(n, "<", ">")
+    n = n.strip()
+    if "::" in n:
+        n = n.rsplit("::", 1)[-1]
+    return n
+
+
+def _base_from_mangled(mangled: str) -> str:
```
tsrikris, line 138: Is this required? How often is this triggered?

TraceLens/TraceUtils/kernel_source/datatypes.py
tsrikris, line 1: Can this be broken down? Just making sure that you need this extra file here.

TraceLens/TraceUtils/kernel_source/contract.py
```diff
@@ -0,0 +1,261 @@
...
+
+import json
+import math
+import os
+import re
+from pathlib import Path
+from typing import Any
+
+#: Bump the major on any field removal or meaning change; consumers gate on it.
+SOURCE_RESOLUTION_SCHEMA_VERSION = "1.0.0"
```
tsrikris: I'd prefer not having this here. Instead log the TraceLens SHA. We'd preferably need only a single version source in the repo

TraceLens/TraceUtils/kernel_source/contract.py
```diff
@@ -0,0 +1,261 @@
...
+    return {
+        "schema_version": SOURCE_RESOLUTION_SCHEMA_VERSION,
+        "generated_by": str(generated_by or ""),
+        "model_name": str(model_name or ""),
+        "framework": str(framework or ""),
+        "entries": list(entries),
+    }
+
+
+def validate_document(doc: Any) -> list[str]:
```
tsrikris: Why is this required here? Ideally if the only producer of this is a script we could just leave it without validating and instead mark unavailable data blank in the JSON

TraceLens/TraceUtils/kernel_source/contract.py
```diff
@@ -0,0 +1,261 @@
...
+                problems.append(
+                    f"entries[{i}] has invalid confidence {confidence!r}; "
+                    "expected a finite number in [0, 1]"
+                )
+    return problems
+
+
+#: A call site is often reported as ``path.py(247): fn_name``; the line and
+#: function ride along in the same string.
+_LINE_SUFFIX_RE = re.compile(
```
tsrikris: Move to the top

TraceLens/TraceUtils/kernel_source/contract.py
```diff
@@ -0,0 +1,261 @@
...
+    if match is None:
+        return text, None, ""
+    return (
+        match.group("path").strip(),
+        int(match.group("line")),
+        str(match.group("function") or "").strip(),
+    )
+
+
+def strip_line_suffix(path: str) -> str:
```
tsrikris: Seems redundant. Can't callers directly just access the data?

TraceLens/TraceUtils/kernel_source/contract.py
```diff
@@ -0,0 +1,261 @@
...
+            return real
+    return ""
+
+
+def path_is_acceptable(path: str, roots: tuple[str, ...]) -> bool:
+    """Whether a rewriting tier may write ``path`` as a resolved location."""
+    return bool(canonical_source_path(path, roots))
+
+
+def read_document(path: Path | str) -> dict[str, Any] | None:
```
tsrikris, line 174: Use the top level dataloader class

TraceLens/TraceUtils/kernel_source/contract.py
```diff
@@ -0,0 +1,261 @@
...
+
+def strip_line_suffix(path: str) -> str:
+    """Return the bare file path from a possibly line-annotated one.
+
+    ``/repo/moe.py(247): _grouped_gemm`` -> ``/repo/moe.py``.
+    """
+    return split_line_suffix(path)[0]
+
+
+def canonical_source_path(path: str, roots: tuple[str, ...]) -> str:
```
tsrikris, line 149: Can you clarify the role of this script?

TraceLens/TraceUtils/kernel_source/validate.py
tsrikris, line 1: Could you clarify why this file exists here? Tied to the comment about the need for validating this Pythonic module, you can:
(1) Move this to kernel_source
(2) Delete it if validation is not required

tsrikris (dismissed):
Additional comments added for clarification

TraceLens/TraceUtils/kernel_source/cli.py
Ahmedhasssan-aig: Some kernels (especially Triton ones) can't be found just from the kernel's name. The extra clue that helps is the "call stack". TraceLens already has the call stack sitting in a column of its report. It hands that list straight to our code through a normal function call. --call-stack-file is useful for manual testing/debugging cases (with no TraceLens report loaded).

TraceLens/TraceUtils/kernel_source/contract.py
Ahmedhasssan-aig: I will drop the local schema version and log the TraceLens SHA instead

TraceLens/TraceUtils/kernel_source/contract.py
Ahmedhasssan-aig: It's a producer self-check and just leave missing data blank in the JSON and skip validation. I can drop runtime validation and just leave missing data blank in the JSON instead

TraceLens/TraceUtils/kernel_source/contract.py
Ahmedhasssan-aig: I will inline it and delete the function

TraceLens/TraceUtils/kernel_source/contract.py
Ahmedhasssan-aig: This is a gate that validates a candidate path before any resolver tier is allowed to record it. It confirms the file truly exists on this host and resolves to somewhere under an allowed search root. This blocks fabricated-but-plausible paths that a mere prefix check would allow. Eventually returns the canonical absolute path on success, otherwise empty string.

TraceLens/TraceUtils/kernel_source/datatypes.py
Ahmedhasssan-aig: This file just holds three small shapes (where the source is, whether it can be edited, and the final result). They sit in their own file so the other parts can use them without pointing back at the resolver. Otherwise, the files would import each other in a loop and break. It's a clean setup, so I'd like to keep it.

TraceLens/TraceUtils/kernel_source/demangle.py
Ahmedhasssan-aig: It's used only when neither itanium-demangler nor c++filt is present. I can not give accurate number but it will be revealed once the code runs in production.

TraceLens/TraceUtils/kernel_source/patchability.py
Ahmedhasssan-aig: We can have a small registry idea as a next step if this grows. Instead of each caller passing its own list, we'd keep one shared list of allowed extensions in a config, and callers add to that. But that's only worth doing if the list is changing a lot but for now, extra_exts covers it.

TraceLens/TraceUtils/kernel_source/index.py
Ahmedhasssan-aig: Regex can do the matching, but the key requirement is that comment removal must not shift byte offsets or newlines, we later map positions to line numbers for pinning. A plain re.sub(..., "") deletes matches and throws those off. The char scanner blanks comments in place (spaces, newlines kept) so positions stay exact.

TraceLens/TraceUtils/kernel_source/index.py
Ahmedhasssan-aig: _PROCESS_INDEX is not an env var. It's the in-memory index cache, kept next to the functions that use it

TraceLens/TraceUtils/kernel_source/validate.py
Ahmedhasssan-aig: I will move it to "kernel_source"

TraceLens/TraceUtils/kernel_source/triton_pin.py
```diff
@@ -0,0 +1,187 @@
...
+    Returns:
+        A :class:`~.datatypes.ResolveResult`. ``method`` is ``"triton_ast"`` (path +
+        pinned line), ``"trace_kernel_file"`` (path only), ``"gate_non_patchable"``
+        (generated Triton), or ``"unresolved"`` (empty/unusable input).
+    """
+    path, line, func = _parse_launcher_form(kernel_file)
+    if not path:
+        return ResolveResult(
+            None, patchable=False, method="unresolved", reason="empty kernel_file"
+        )
```
kyle-hoffmeyer, line 208: Is `kernel_file` something collected via profiling? Is it a field in the trace?

Currently, `resolve_triton_source` is the main method of resolving triton kernels, but it requires the `kernel_file` field. If this field doesn't exist, the default path is called with `TraceLens_resolve_kernel_source --kernel`. But this  doesn't scan through .py files. Just want to confirm that `kernel_file` is always present for triton kernels, because otherwise triton kernel source resolution fails.

TraceLens/TraceUtils/kernel_source/editable.py
```diff
@@ -0,0 +1,78 @@
...
+            Case-insensitive; a leading dot is optional.
+
+    Returns:
+        ``True`` for native device code or a repo-resident Triton ``.py``;
+        ``False`` for empty paths, generated Triton, and non-source files.
+    """
+    if not path:
+        return False
+    native_exts = _NATIVE_SOURCE_EXTS
+    if extra_exts:
```
kyle-hoffmeyer, line 57: `is_editable_source` is only called in `triton_pin.py` and `extra_exts` is not called. Therefore, the extra_exts parameter is dead code. Is it necessary?

TraceLens/TraceUtils/kernel_source/editable.py
```diff
@@ -0,0 +1,78 @@
...
+        # Normalize each extra ext to lowercase with a leading dot before adding.
+        extra = tuple(
+            e.lower() if e.startswith(".") else "." + e.lower() for e in extra_exts
+        )
+        native_exts = native_exts + extra
+    low = path.lower()
+    if low.endswith(native_exts):
+        return True
+    if low.endswith(".py"):
+        if kind == "triton_inductor_generated":
```
kyle-hoffmeyer: `kind` is also never specified so this path is never taken.

TraceLens/TraceUtils/kernel_source/triton_pin.py
```diff
@@ -0,0 +1,187 @@
...
+        line_str = match.group("pline") or match.group("cline")
+        return (
+            match.group("path").strip(),
+            int(line_str) if line_str else None,
+            match.group("func") or "",
+        )
+    return text, None, ""
+
+
+def editable_trace_source(kernel_file: str, kind: str = "") -> str:
```
kyle-hoffmeyer: Is this wrapper function needed?

TraceLens/TraceUtils/kernel_source/cli.py
```diff
@@ -0,0 +1,137 @@
...
+    if args.triton_kernel_file:
+        result = resolve_triton_source(args.triton_kernel_file, symbol=args.kernel)
+        print(json.dumps(_result_to_dict(result), indent=2))
+        return 0
+
+    if not args.kernel:
+        print("error: --kernel (or --triton-kernel-file) is required", file=sys.stderr)
+        return 2
+
+    call_stack = _read_call_stack(args.call_stack_file)
```
kyle-hoffmeyer: Can we add documentation for this flag in docstring at the top of this file?

TraceLens/TraceUtils/kernel_source/cli.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if args.gate_only:
```
kyle-hoffmeyer: Is this `gate_only` flag needed? classify_patchability is ran either way in `resolve`.

TraceLens/TraceUtils/kernel_source/resolver.py
```diff
@@ -0,0 +1,138 @@
...
+    search_paths: Sequence[str | Path] | None = None,
+    *,
+    op_name: str = "",
+    call_stack: Sequence[str] = (),
+    run_gate: bool = True,
+    index_obj: index.SourceIndex | None = None,
+) -> ResolveResult:
+    """Run the cheap patchability gate first, then resolve survivors via :func:`resolve_source_path`."""
+    if run_gate:
+        gate = classify_patchability(
```
kyle-hoffmeyer: Why is this behind a flag `run_gate`? Shouldn't this always be ran?

TraceLens/TraceUtils/kernel_source/editable.py
```diff
@@ -0,0 +1,78 @@
...
+_NATIVE_SOURCE_EXTS = (".cu", ".cuh", ".hip", ".h", ".hpp")
+
+# Path markers for compiler-generated Triton (torch.compile / inductor). These
+# files are produced at compile time and have no durable source to rewrite.
+# Covers both the classic inductor temp dir and vLLM's on-disk compile cache
+# (e.g. ``~/.cache/vllm/torch_compile_cache/.../inductor_cache/....py``).
+_GENERATED_MARKERS = ("torchinductor", "inductor_cache", "torch_compile_cache")
+
+
+def is_editable_source(
```
kyle-hoffmeyer, line 36: Could we somehow move the logic in this function into `classify_patchability`? And just have `classify_patchability` ran before `resolve_triton_source` and `resolve` in cli.py?

TraceLens/TraceUtils/kernel_source/resolver.py
kyle-hoffmeyer: Moving `classify_patchability` to the beginning of main() in cli.py will make this wrapper function no logner required and `resolve_source_path` can be called directly

TraceLens/TraceUtils/kernel_source/editable.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+
+__all__ = ["is_editable_source"]
+
+# Native device-code extensions that are always editable.
+_NATIVE_SOURCE_EXTS = (".cu", ".cuh", ".hip", ".h", ".hpp")
```
kyle-hoffmeyer: This same list is defined in index.py

TraceLens/TraceUtils/kernel_source/demangle.py
```diff
@@ -0,0 +1,211 @@
+###############################################################################
+# Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""Turn a raw GPU kernel name from a trace into its bare name.
```
kyle-hoffmeyer, line 7: Did you have a chance to verify itanium-demangler efficacy vs. c++filt? Once we verify itanium-demangler works fine, we can remove the c++filt fallback and remove a lot of code

TraceLens/TraceUtils/kernel_source/resolver.py
```diff
@@ -0,0 +1,138 @@
...
+from .editable import is_editable_source
+from .datatypes import ResolveResult, SourceLocation
+from .patchability import classify_patchability
+
+log = logging.getLogger(__name__)
+
+__all__ = ["resolve_source_path", "resolve"]
+
+# Framework labels inferred from a resolved path, for SourceLocation.framework.
+_FRAMEWORK_HINTS = ("aiter", "sglang", "vllm")
```
kyle-hoffmeyer: This list is defined in index.py

kyle-hoffmeyer (commented):
Added some comments

TraceLens/TraceUtils/kernel_source/triton_pin.py
Ahmedhasssan-aig: Yes, kernel_file is a trace field (from the Triton launch args, cpu_op["args"]["kernel_file"]), not something we compute.  On PyTorch >2.4 it's always present and resolve_triton_source is the primary path for it. If needed, I can add a fallback that uses the python call-stack frames or an AST name-scan to find .py, in case it's missing.

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: I added extra_exts specifically because Tharun asked for a way to extend the editable-file list in the future without editing this module.

TraceLens/TraceUtils/kernel_source/demangle.py
Ahmedhasssan-aig: Yes, I tested it across 5,826+ symbols. I added some tests related to this verification.

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: kind is there if someone wires the kernel's type from the trace into the resolver. Currently, it's not being used so we can drop it.

TraceLens/TraceUtils/kernel_source/triton_pin.py
Ahmedhasssan-aig: It's only used once internally. I'll inline it into resolve_triton_source and drop it from the public API

TraceLens/TraceUtils/kernel_source/cli.py
Ahmedhasssan-aig: The gate_only branch just returns the gate verdict and stops early, skipping the expensive resolve step (index build + file scan). classify_patchability does run inside resolve too, but resolve does the full resolution first. This is a cheap "verdict only" path.

TraceLens/TraceUtils/kernel_source/resolver.py
Ahmedhasssan-aig: By default, it always runs (run_gate=True). The flag is just an escape, if a caller already ran the gate earlier, they could set run_gate=False to avoid doing it twice. So, it's on by default and only skipped when it'd be redundant.

Ahmedhasssan-aig: I'd keep resolve() in the library. It's the one-call "gate + resolve" entry for integrated callers and the validation tool. Without it, every non-CLI caller has to redo the gate-then-resolve steps themselves. But I'll run the gate in main() and call resolve_source_path directly, so the CLI no longer needs the wrapper.

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: They run at different stages on different inputs. classify_patchability runs on the kernel name up front, before any file is known. is_editable_source runs on each resolved file path inside the finder loop (to reject generated//tmp files). They cannot merge cleanly.

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: _NATIVE_SOURCE_EXTS in index.py is "which file types to scan" and in editable.py is "which count as editable". So, they can diverge.

TraceLens/TraceUtils/kernel_source/resolver.py
Ahmedhasssan-aig: I'll drop _FRAMEWORK_HINTS and have _framework_of iterate over index._known

TraceLens/TraceUtils/kernel_source/demangle.py
tsrikris: I'd remove the c++filt code completely

Ahmedhasssan-aig: Done

.coveragerc
```diff
@@ -6,6 +6,9 @@ omit =
     TraceLens/PerfModel/benchmarking/*
     TraceLens/PerfModel/origami_helper.py
     TraceLens/PerfModel/run_perf_model.py
+    # Manual validation script: exercises the resolver against a real framework
```
tsrikris: Remove in-line comment

TraceLens/TraceUtils/kernel_source/cli.py
tsrikris: Can you clarify then why we need a separate file? Meaning can we not pass it in as a usual CLI arg, i.e --call-stack 'nn.Module, Kernel_Launcher, Kernel' and so on?

TraceLens/TraceUtils/kernel_source/index.py
```diff
@@ -0,0 +1,738 @@
...
+    for dirpath, _dirs, names in os.walk(root):
+        for nm in names:
+            if nm.lower().endswith(_NATIVE_EXTS):
+                yield Path(dirpath) / nm
+
+
+# --- Triton kernel-definition scanning (for the stage-2 .py fallback) --------
+# Decorators that mark a Triton device-kernel def (``@triton.jit`` / ``@jit``
+# and the autotune/heuristics wrappers stacked on top of a jit'd kernel).
+_TRITON_DECORATORS = frozenset({"jit", "autotune", "heuristics"})
```
tsrikris: Move to the top

TraceLens/TraceUtils/kernel_source/validate.py
tsrikris: Revisiting this comment. Are you planning to keep this still?

tsrikris (commented):
Few pending comments need addressing. Mostly looks good!!!

TraceLens/TraceUtils/kernel_source/cli.py
kyle-hoffmeyer: Bumping this

TraceLens/TraceUtils/kernel_source/editable.py
kyle-hoffmeyer: Shouldn't we only be scanning files we can edit though? Are they not the same thing?

TraceLens/TraceUtils/kernel_source/editable.py
kyle-hoffmeyer: This seems to still be here

TraceLens/TraceUtils/kernel_source/editable.py
kyle-hoffmeyer: I think adding an optional `file_path` parameter to `classify_patchability` would allow the merging of these two functions. I think this would be simpler as the logic for determining whether a kernel is patchable would be contained in one file and we could delete this file as this is the only function in this file.

TraceLens/TraceUtils/kernel_source/resolver.py
kyle-hoffmeyer: Ok, if you think it's useful we can leave it. I just want to remove flags that won't be used in production.

kyle-hoffmeyer: The `resolve` function seems to still be here

TraceLens/TraceUtils/kernel_source/triton_pin.py
kyle-hoffmeyer: Ok just wanted to check. Thanks

kyle-hoffmeyer (commented):
Some previous comments are unaddressed

kyle-hoffmeyer (commented):
Some previous comments are unaddressed

TraceLens/TraceUtils/kernel_source/validate.py
Ahmedhasssan-aig: We can keep the validation function for now. Additionally, I need to conduct an apple-to-apple comparison of the Magpie kernel source mapping with ours. This script will be helpful there too.

TraceLens/TraceUtils/kernel_source/cli.py
Ahmedhasssan-aig: The real frames are not short labels like 'nn.Module, Kernel_Launcher, Kernel' instead they're free-form strings carrying file paths like ""/root/.cache/.../torchinductor_abc/xyz.py", line 12, in triton_poi_fused_0". Quoting this on a single command would be tricky.

TraceLens/TraceUtils/kernel_source/cli.py
Ahmedhasssan-aig: The --call-stack-file flag is now documented in the module docstring at the top of cli.py the "call-stack frames" example (lines 15–20). It explains the one-frame-per-line format and notes it's a cli only convenience

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: I have unified it

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: I would prefer to keep these as two functions. They look similar but operate at different stages on different inputs

TraceLens/TraceUtils/kernel_source/editable.py
Ahmedhasssan-aig: Yes, cursor messed up with my previous edits. I have removed it now

TraceLens/TraceUtils/kernel_source/resolver.py
Ahmedhasssan-aig: resolve() is the library's single "gate + resolve" entry point, and it's not cli only. Validation steps calls it in two places (validate.py:93, validate.py:163). Like I said, I will keep it

TraceLens/TraceUtils/kernel_source/validate.py
tsrikris: I'm still not keen on keeping validation code in a production repo. I'd either suggest removing this and rely on a local copy for short term validation; or open a follow up ticket to clean up. Also is this validation invoked in the main flow? Validating the output of output pythonic modules with another piece of code is not efficient.

Ahmedhasssan-aig: This is not invoked in the main flow. It's used on-demand just to test different modules of the actual code. I will drop it from the production code and keep a local copy for testing.

Ahmedhasssan-aig:
@tsrikris @kyle-hoffmeyer 
If you have no further comments. Can we approve this PR?

TraceLens/TraceUtils/kernel_source/cli.py
tsrikris: I'm still not convinced of the file as an input. A user that's going to use this file format is going to have to convert the CSV row into a text file that's sent into this. I think you have to somehow factor in that the cleanest approach is going to involve reading the perf CSV as a dataframe and directly pass that in.

kyle-hoffmeyer (approved):

TraceLens/TraceUtils/kernel_source/cli.py
Ahmedhasssan-aig: Droping --call-stack-file flag for now. We can add it back if we find the need of stack-based Inductor check from a shell.

tsrikris (approved):

tsrikris:
@Ahmedhasssan-aig reminder to open any pending issues or optimizations in follow up tickets. A documentation README there along with a page under docs would be useful

PR #996, opened 2026-09-02
Fix CPU-GPU linking when ac2g start events are missing.

devalshahamd:
This pull request improves the robustness of the `_find_corresponding_output_event` method in `trace_to_tree.py` by adding a fallback mechanism for runtimes that emit only partial ac2g flow events. The method now attempts a direct correlation-id match when the usual start/end event pairing is incomplete.

**Enhanced output event matching logic:**

* Added a fallback to match output events by unambiguous correlation-id if only the finish half of the ac2g flow is present, improving compatibility with runtimes like `hipDrvLaunchKernelEx`.
* Refactored the logic to only require both ac2g start and end events for the primary matching path, and otherwise use the fallback correlation-id matching.

# Pull Request Template

> **Note to AMDers:**  

### DIFF
--- a/TraceLens/Trace2Tree/trace_to_tree.py
+++ b/TraceLens/Trace2Tree/trace_to_tree.py
@@ -1199,26 +1199,25 @@ class TraceToTree(BaseTraceToTree):
         # 1. Get the linking id from the input event
         # 2. Find the corresponding start and end ac2g events for the linking id
         # 3. Find the output event using the pid, tid, and linking id of the end ac2g event
+        # 4. Some runtimes (e.g. hipDrvLaunchKernelEx) emit only the finish half of the
+        #    ac2g flow, so fall back to an unambiguous correlation-id match
         link_id = input_event.get(TraceEventUtils.TraceKeys.Args, {}).get(
             self.linking_key
         )
         ac2g_start_event = self.ac2g_event_map["start"].get(link_id)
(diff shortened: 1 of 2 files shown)


TraceLens/Trace2Tree/trace_to_tree.py
```diff
@@ -1199,26 +1199,25 @@ def _find_corresponding_output_event(self, input_event):
...
-
-        output_event = self.pid_tid_event_map.get((pid, tid, link_id))
+        output_event = None
+        if ac2g_start_event and ac2g_end_event:
+            pid = ac2g_end_event.get(TraceEventUtils.TraceKeys.PID)
+            tid = ac2g_end_event.get(TraceEventUtils.TraceKeys.TID)
+            end_link_id = ac2g_end_event.get("id")
+            output_event = self.pid_tid_event_map.get((pid, tid, end_link_id))
+
+        if output_event is None:
```
gabeweisz: Quibble: could be more readable as an else: to line 1211

gabeweisz (approved):

TraceLens/Trace2Tree/trace_to_tree.py
devalshahamd: Thanks for the suggestion! I included the return statement in the if condition, so fewer number of lines now and I think it is more readable.

gabeweisz (approved):

PR #1000, opened 2026-09-02
Enable `AMD-AGI/TraceLens` to reproduce minimal tests required by `amd/skills`

danielholanda:
# Description

Enable `AMD-AGI/TraceLens` to reproduce minimal tests required by `amd/skills`

## Details

**This PR:**
* Moves the evals required by AMD Skills Catalog from the catalog to the source repo(`AMD-AGI/TraceLens` becomes the single source of truth).
* Adds a dispatchable workflow file that enables `AMD-AGI/TraceLens` to run this minimal set of tests easily to validate functionality before merging into the AMD-skills repo.

Note: This PR also ensures that the skill name and the folder name match each other. This is in general a good practice as some tools do expect those names to match. Please also note that this could also be solved by changing the folder name instead of the skill name. When importing it to `AMD Skills` we are still importing it as `tracelens-analysis-orchestrator`.

With those changes, `AMD-AGI/TraceLens` becomes the single source of truth for the `tracelens-analysis-orchestrator` skill.


TraceLens/Agent/Analysis/skills/analysis-orchestrator/SKILL.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: tracelens-analysis-orchestrator
```
danielholanda, line 2: Same comment as in the PR description:
> This PR also ensures that the skill name and the folder name match each other. This is in general a good practice as some tools do expect those names to match. Please also note that this could also be solved by changing the folder name instead of the skill name. When importing it to AMD Skills we are still importing it as tracelens-analysis-orchestrator.

TraceLens/Agent/Analysis/skills/analysis-orchestrator/evals/evals.json
```diff
@@ -0,0 +1,50 @@
+{
```
Ahmedhasssan-aig: @danielholanda Why do we need evals.json, and where exactly is it used?

danielholanda: @Ahmedhasssan-aig The AMD skills catalog requires certain evals to be present for every single skill that joins the catalog.  AMD skills imports `tracelens` every time it gets updated and those tests run during this process. Happy to provide additional info offline if needed.

tsrikris: @Ahmedhasssan-aig I think this file is a key component of the AMD skills repo and we will not be running it ourselves

tsrikris (approved):
LGTM

Ahmedhasssan-aig (approved):
LGTM!

Issue #872, opened 2026-08-01
[Feature Request] Add roofline/architecture support for Radeon AI PRO R9700S (gfx1201) and Ryzen AI Halo Box (gfx1151)


xiaofei-zheng:
# [Feature Request] Add roofline/architecture support for Radeon AI PRO R9700S and Ryzen AI Halo Box

## 1. Background

TraceLens currently ships architecture performance specs only for data-center GPUs:

- `TraceLens/Agent/Analysis/utils/arch/MI300X.json`
- `TraceLens/Agent/Analysis/utils/arch/MI325X.json`

We are expanding AI inference profiling to **client/edge AMD platforms** that are increasingly used for local model deployment:

| Platform | Product | LLVM Target | Architecture | Memory | Peak BW |
|----------|---------|-------------|--------------|--------|---------|
| **Discrete GPU** | Radeon AI PRO R9700S | `gfx1201` | RDNA 4 (Navi 48) | 32 GB GDDR6 | 640 GB/s |
| **APU (Halo Box)** | Ryzen AI Max+ 395 (Radeon 8060S) | `gfx1151` | RDNA 3.5 (Strix Halo) | Up to 128 GB unified LPDDR5x | ~256 GB/s |

Without correct architecture specs, roofline analysis, TFLOP/s / TB/s efficiency metrics, and Origami simulation produce **incorrect or missing bounds** on these platforms — making TraceLens reports unreliable for the workloads we profile there (including HunYuan world models, Stable Diffusion, Flux, and LLMs).

## 2. Symptoms

When profiling on R9700S or Ryzen AI Halo Box:

- `--gpu_arch_json_path` has no bundled spec for `gfx1201` or `gfx1151`; users must manually craft JSON or fall back to MI300X specs (wildly incorrect roofline bounds)
- Roofline efficiency percentages are meaningless (e.g., a kernel at 10 TFLOP/s might show 2% on MI300X roofline but 20% on R9700S roofline)
- TraceLens Agent optimization reports reference wrong hardware headroom
- Origami simulation (`enable_origami`) cannot select the correct GPU architecture model
- Auto-detection from trace metadata does not map `gfx1201` / `gfx1151` to architecture specs

### 3.1 Radeon AI PRO R9700S (`gfx1201`, RDNA 4)

| Spec | Value | Source |
|------|-------|--------|
[...]

PR #1011, opened 2026-09-09
Add Radeon 8060S (gfx1151) arch spec

mgehre-amd:
Adds a bundled arch spec for the Radeon 8060S integrated GPU (RDNA 3.5, gfx1151) so roofline analysis works on Ryzen AI Max+.
Data is taken from public sources such as https://gpuopen.com/learn/wmma_on_rdna3/.

Allow spec files without `memory_gb` field. Specs for APU parts like Strix Halo won't have a fixed `memory_gb` because those use unified memory (sharing with host DDR) that is configurable in the BIOS.


TraceLens/Agent/Analysis/utils/arch/Radeon_8060S.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Radeon 8060S",
+    "gfx_arch": "gfx1151",
+    "memory_type": "unified",
```
tsrikris: Could you pls clarify how you plan to use the memory_type variable?

TraceLens/Agent/Analysis/utils/arch/Radeon_8060S.json
```diff
@@ -0,0 +1,16 @@
...
+    "max_achievable_tflops": {
+        "matrix_fp16": 59.4,
+        "matrix_bf16": 59.4,
+        "matrix_fp32": 29.7,
+        "matrix_int8": 59.4,
+        "vector_fp16": 29.7,
+        "vector_bf16": 29.7,
+        "vector_fp32": 29.7
+    },
+    "_reference": "https://gpuopen.com/learn/wmma_on_rdna3/"
```
tsrikris: On the _reference field, would you recommend having this in place? The other values are benchmarked MAF values

TraceLens/Agent/Analysis/utils/arch/Radeon_8060S.json
mgehre-amd: You are right, let me remove that until it has a proper use. I saw that https://github.com/AMD-AGI/TraceLens/issues/872 had mentioned it, but it doesn't make sense to add it without user.

TraceLens/Agent/Analysis/utils/arch/Radeon_8060S.json
mgehre-amd: I fine to remove the _reference field, the origin is also in the commit message. I got publicly available data to get started. To proper benchmark MAF values, I want to modify the microbenchmark (different shapes achieve MAF on the smaller Strix Halo).

tsrikris (approved):

tsrikris (approved):

PR #1012, opened 2026-09-09
Create script to make new reference archives new references

gabeweisz:
New archives as of 09/09/2026

# Pull Request Template

> **Note to AMDers:**  


tsrikris (approved):

tsrikris:
@Ahmedhasssan-aig  can you review?

agent_evals/Analysis/eval_scripts/pack_test_archives.sh
```diff
@@ -0,0 +1,149 @@
...
+    fi
+
+    echo "Packing $rel -> $archive"
+
+    tar -czf "$archive" \
+        --exclude='analysis_output' \
+        --exclude='analysis_stream.ndjson' \
+        --exclude='__pycache__' \
+        -C "$REPO_ROOT" \
+        "$rel"
```
Ahmedhasssan-aig, line 124: Sometimes, I see a few redundant .csv file generated in the directory. Can we add an additional check at the end of the script to verify if the generated pack is correct?

Ahmedhasssan-aig:
Just one suggestion, otherwise looks good to me!

agent_evals/Analysis/eval_scripts/pack_test_archives.sh
gabeweisz: Can you be more specific? What does the check look like - how do we know what files are needed and what files are not?

Ahmedhasssan-aig: Like a small function/script at the end of the pack generation that unzip the generated archive and verifies, it has analysis_output_ref dirtory (containing only perf_report_csv directory and analysis.md) and a trace file. If there is anything redundant, it should be flagged/removed

tsrikris: I don't think we should be verifying a pythonic/shell script's output with another script, just adds bloat to the repo. I'd instead make sure the list of folders that are tarred contain what's needed.

gabeweisz: @Ahmedhasssan-aig how do we know what is redundant? We can add a specific list of files instead of * if you tell us how to create that list of files

Ahmedhasssan-aig: Yes, I agree. What I meant was that anything other than the perf_report_csv directory, analysis.md, and the json.gz file should be removed. We can add a check and validate during the tar file creation. Previously, when I was doing it manually with cursor, it used to push some duplicate csv files to main analysis_output_ref directory. So, I was runing another check on generated tar file before raising the PR to the TraceLens

tsrikris: I'd say to close this out, the script can consume any files we don't want to be added to the tar file. If there's a specific list we can add that in, but we shouldn't be running scripted tests for scripts (outside of CI). If there's anything you'd like to add @Ahmedhasssan-aig pls do so if not let's close this merge off.

Ahmedhasssan-aig (approved):

PR #1022, opened 2026-09-10
Inference split: keep host ops between step annotations

### LINKED ISSUE #1023: Inference split drops vLLM compute_logits / lm_head after execute_* envelope
## Summary
Inference-trace split (`extract_iteration`) keeps CPU ops only while they are nested inside each iteration annotation (`ts` .. `ts+dur`). On some vLLM serving traces, `sample_tokens` / `compute_logits` / the vocab (lm_head) GEMM run **after** that envelope and **before** the next `execute_*` annotation. Those ops never land in the extracted window or GEMM tables, even though they are shaped GEMMs in the unsplit chrome (`vllm::rocm_unquantized_gemm` with Input Dims filled). This is a windowing bug, not a missing perf-model bind.

## Scope (not every model)
The failure mode is **vLLM post-annotation sample**, not “all models lack lm_head.”

- **Hits:** dense TP=1 serving where lm_head sits outside `execute_*` — observed on Qwen3-0.6B / 8B / 14B and Llama-3.1-8B (body GEMMs present in the split report, vocab N absent).
- **Does not always hit:** other vLLM configs in the same serving corpus still show vocab-sized `vllm::rocm_unquantized_gemm` in split reports (lm_head still inside the kept window).
- **SGLang is a different hole:** lm_head is often inside the CUDA graph; serving-only reports can show zero GEMM rows until `--capture_folder` merge. Do not treat that as this splitter bug.

## Fix
Keep CPU ops from each root’s start until the **next same-pattern iteration annotation** on that thread. GPU kernels still attach by correlation. Neighboring steps must not leak in.

The last step of a profile (no following annotation) still ends at `ts+dur`, so that one lm_head can still drop.

## PR
#1022

ajassani:
## Summary
- Split windows previously kept CPU ops only while they were nested inside each iteration annotation (`ts` .. `ts+dur`). Serving stacks launch work after that envelope and before the next step — vLLM `compute_logits` / lm_head GEMMs are the usual example — so those ops never made it into the extracted trace or GEMM tables.
- Each root's CPU window now runs until the next same-pattern iteration annotation on that thread. GPU kernels still attach by correlation, so they can sit outside the host window. Neighboring steps do not leak in.
- Not every model: this bites dense TP=1 vLLM serving where sample sits outside `execute_*` (Qwen3-0.6B/8B/14B, Llama-3.1-8B). Other vLLM configs can still keep vocab GEMMs in the split report. SGLang in-graph lm_head is a separate `--capture_folder` issue.

Fixes #1023

## Test plan
- [x] `pytest tests/test_split_inference_trace_annotation.py`
- [ ] Re-split a vLLM serving trace and confirm `compute_logits` / vocab-sized GEMMs appear in the extracted window and GEMM sheet
- [ ] Spot-check an SGLang serving split still isolates `step[DECODE|EXTEND]` windows

### DIFF
--- a/TraceLens/TraceUtils/split_inference/trace_extraction.py
+++ b/TraceLens/TraceUtils/split_inference/trace_extraction.py
@@ -13,6 +13,7 @@ import zipfile
 
 from tqdm import tqdm
 
+from ...util import TraceEventUtils
 from ..annotation_utils import (
     ITERATION_BACKUP_PATTERNS,
     ITERATION_PATTERNS,
@@ -22,6 +23,8 @@ from ..annotation_utils import (
     is_decode_only,
     iteration_details,
 )
+
+_K = TraceEventUtils.TraceKeys
 
 
 def _annotation_pattern(name: str):
@@ -41,20 +44,20 @@ def _next_same_pattern_ts(root: dict, events: list[dict]) -> float | None:
     duration drops that work. Bounding by the next matching annotation keeps it
     without leaking the following step's nested ops.
     """
-    pattern = _annotation_pattern(root.get("name", ""))
+    pattern = _annotation_pattern(root.get(_K.Name, ""))
     if pattern is None:
         return None
-    iter_ts = root.get("ts", 0)
-    iter_tid = root.get("tid")
-    iter_pid = root.get("pid")
+    iter_ts = root.get(_K.TimeStamp, 0)
+    iter_tid = root.get(_K.TID)
+    iter_pid = root.get(_K.PID)
     next_ts = None
     for e in events:
-        ts = e.get("ts")
+        ts = e.get(_K.TimeStamp)
         if ts is None or ts <= iter_ts:
             continue
-        if e.get("tid") != iter_tid or e.get("pid") != iter_pid:
+        if e.get(_K.TID) != iter_tid or e.get(_K.PID) != iter_pid:
             continue
-        if pattern.match(e.get("name") or ""):
+        if pattern.match(e.get(_K.Name) or ""):
             if next_ts is None or ts < next_ts:
                 next_ts = ts
     return next_ts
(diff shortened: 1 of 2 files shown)


TraceLens/TraceUtils/split_inference/trace_extraction.py
```diff
@@ -23,6 +23,52 @@
...
+def _next_same_pattern_ts(root: dict, events: list[dict]) -> float | None:
+    """Timestamp of the next same-pattern iteration annotation on this thread.
+
+    Serving runtimes (vLLM ``compute_logits``, SGLang ``_compute_lm_head``) often
+    launch work *after* the step annotation's duration ends and *before* the
+    next step annotation starts. Bounding the CPU window by the annotation
+    duration drops that work. Bounding by the next matching annotation keeps it
+    without leaking the following step's nested ops.
+    """
+    pattern = _annotation_pattern(root.get("name", ""))
```
gabeweisz: We should use our defined constants for field names instead of duplicating the text here

Defined constants are in https://github.com/AMD-AGI/TraceLens/blob/4447ce9f27070f3f31322e9dd6052a34b03c4e23/TraceLens/util.py#L624

TraceLens/TraceUtils/split_inference/trace_extraction.py
```diff
@@ -23,6 +23,52 @@
...
+
+
+def _cpu_window_end(root: dict, next_sibling_ts: float | None, events: list[dict]):
+    """Exclusive end of the CPU window for one iteration root."""
+    if next_sibling_ts is not None:
+        return next_sibling_ts
+    next_ts = _next_same_pattern_ts(root, events)
+    if next_ts is not None:
+        return next_ts
+    return root.get("ts", 0) + root.get("dur", 0)
```
gabeweisz: Defined constants

TraceLens/TraceUtils/split_inference/trace_extraction.py
```diff
@@ -86,10 +137,16 @@ def extract_iteration(
     # Compute the global time window for all iteration roots
     if not iteration_roots:
         return trace_json.copy(), [], 0, 0, 0
+    roots_by_ts = sorted(iteration_roots, key=lambda r: r.get("ts", 0))
```
gabeweisz: Defined constants

gabeweisz (changes requested):

TraceLens/TraceUtils/split_inference/trace_extraction.py
ajassani: Done in de202a6c — new window helpers and the adjacent root tid/pid/ts reads use TraceEventUtils.TraceKeys. The pre-existing CPU pre-filter loop in this function still uses chrome string keys, same as the rest of the file.

gabeweisz (approved):

PR #1024, opened 2026-09-11
Kernel shape profiler registry

mohbasit:
Currently to get shapes in SGLang runs, we have been patching the docker image, this method removes the patching and this tool can be used to get shapes in traces for SGlang framework by just putting the file in the PYTHONPATH and enabling flags.


gabeweisz:
Looks good to me but checks fail and this needs tests

mohbasit:
@gabeweisz added tests, they have been skipped on absence of torch due to the temporary nature of the module for now. Will transition to a more generic module in the future.

TraceLens/TraceUtils/kernel_shape_tool/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Launchers come from an explicit registry plus auto-discovery, so no
+serving-framework source is patched.
+
+## Files
+
+| File | Role |
+|------|------|
+| `kernel_shape_profiler.py` | Wraps launcher functions as custom ops and rebinds every module-level reference to them. |
+| `sitecustomize.py` | Auto-loaded shim that drives `enable()` / `disable()` from the torch-profiler window so nothing is wrapped outside a profiling run. |
```
devalshahamd, line 34: Is the module autoloaded for all torch-profile instances or only when shape recording is enabled in  PyTorch profiler?

devalshahamd (commented):
I have a doubt on when the shape profiler is enabled. We should make sure that it is enabled and loaded only when torch profiler instance has record shapes enabled. Otherwise LGTM.

TraceLens/TraceUtils/kernel_shape_tool/README.md
mohbasit: it is only loaded when the Pythonpath is set, 

export PYTHONPATH=/path/to/kernel_shape_tool:$PYTHONPATH
export TRACELENS_SHAPE_DISCOVERY=1

both these env variables need to be set, only then the module is loaded

devalshahamd (approved):

PR #1032, opened 2026-09-15
fix(packaging): make JAX XPlane dependencies optional

iraj465:
TraceLens requires `xprof==2.20.1` for every installation, but that release has no wheel for Python 3.13 or 3.14. Move xprof and its protobuf constraint into the `jax` extra so PyTorch and rocprof analysis can install without the JAX converter.

JAX XPlane users must now install `TraceLens[jax]` with an interpreter supported by xprof. Both protobuf entrypoints retain the existing xprof-to-tensorboard fallback and report how to install the extra when neither converter is available. The JAX dependency pin is preserved; documentation and the unit-test workflow explicitly install the extra.

Validation:

- Merged source import, JSON and gzip parsing, and four report CLI help commands pass on Python 3.14.7 with JAX/protobuf converter imports blocked.
- Python 3.14 wheels-only base dependency resolution and built-wheel metadata checks pass; xprof/protobuf are selected only for the JAX extra.
- Focused package/util, kernel-source, and microbench tests: 297 passed, 22 skipped, 3 GPU tests deselected.
- Complete non-GPU suite on the merged candidate with Python 3.12: 3,070 passed, 42 skipped (306 seconds). The existing skips cover optional Torch/Origami dependencies and unavailable fixtures; this host lacks the HIP development SDK needed to build Origami.
- Black, selected Ruff rules, and the import graph contract pass. No reference outputs changed.


gabeweisz:
How much would it simplify things to drop support for Jax older than 0.8 or 0.9? Jax 0.6 was the version that needed a very special version of xprof

gabeweisz:
I want to resolve this versioning issue, but we need to support newer Pytorch for ROCm 10 - so we can't wait. Making Jax optional is the best option right now

gabeweisz (approved):

PR #1044, opened 2026-09-17
TraceLens Agent JSON Post-Processing

tsrikris:
## Summary

This adds a deterministic renderer that turns a finished `analysis.md` into a typed,
structured `analysis.json`, so downstream consumers can read a contract instead of scraping
markdown. `render_analysis_json(path_to_analysis_md)` writes the JSON next to the md. It
covers the compute tier only for v1. The producer parses, it does not decide. Every value it
emits is read straight from the markdown the analysis agent already wrote, and it assumes
that markdown is contract-valid, since the same pipeline validates it upstream before this
step runs. V1 is only meant for the standalone analysis mode for Hyperloom consumption.

The same md in always gives byte-identical JSON out. That was checked across the real corpus
reports, both in-process and cross-process.

## What changed

- **New `post_processing` package.** Three files: `analysis_json.py` (the renderer),
  `candidate_schema.py` (the typed contract), and `__init__.py` (which re-exports
  `render_analysis_json` and the TypedDicts).
- **Two parser classes, no loose helpers.** `AnalysisMdParser` reads the md (title, report
  info, exec summary, top operations, findings, appendix). `ComputeGrouper` turns
  data-table rows into operation-grouped tasks. Every cell and format helper is a
  `@staticmethod` on the class that uses it, so there is nothing module-level that only one
  class calls.
- **Array-of-structs grouping.** One task per operation, and `members[]` holds one member
  per data-table row. A fan-out row (several kernels in one row) stays a single member whose
  `kernel_name` is a `list[str]`. Grouping keys on the raw `Operation` cell on the agentic
  path, or on `kernel_name[0]` on the deterministic-fallback path, where the Operation cell
  only repeats the kernel name. A `—` Operation cell becomes `null`, not the raw glyph.
- **Impact from markers, with a currency-correct fallback.** Each member's `impact_score`
  reads the `kind=op_row impacts=CSV` marker by row position. When that value is missing,
  either because the block has no `op_row` marker or because the row's CSV entry is the
  spec-legal `—` null, the score falls back to the row's `pct_e2e` scaled by
  `HEURISTIC_FRACTION_MID` (0.30). That is the same fraction the impact ladder
  (`analysis_utils._row_impact_score`) and the deterministic fallback writer already use, so
  a fallback score is comparable to a real one instead of being roughly 3.3x too high. A
  null `pct_e2e` gives `0.0`. The task `impact.mid` is the member sum, and `low`/`high` sum
  the finding-level markers.
- **`library` derived from the P-item heading.** Each member carries the backend token
  (`AITER`, `Tensile`, `PyTorch Native`, `MIOpen`) parsed from the trailing `(<Library>)`
  suffix on the compute heading. This is the same signal the analysis agent is required to
  emit (`sub_agent_spec.md`) and the same one the downstream consumer already derives for
  itself. It is `None` when the heading has no paren, which is the spec's "Unknown" case. It
  is stamped per block onto every member, the same way `category` is.
- **`analysis_md_rank` per member, plus `[P<rank>]` prose labels.** Each member carries
  `analysis_md_rank = P{rank}`, read from its `<!-- reasoning-candidate tier=compute rank=N -->`
  marker, so a JSON row points back to its source P-item in the md. When a task merges more
  than one finding, the joined `identification`, `reasoning`, and `resolution` prose is
  prefixed `[P1] ... [P2] ...` per source block. A single-finding task is left unprefixed.
- **One marker regex, one owner.** The `op_row` impacts CSV pattern lives on
  `MarkerValidator.OP_ROW_IMPACTS_RE` and is used by both the validator and this renderer,
  so the writer and reader cannot drift apart.
- **Prose caps unified at 2000.** `IDENTIFICATION_CAP`, `REASONING_CAP`, and
  `RESOLUTION_CAP` are all 2000. Over-cap prose drops whole trailing blocks and sets
  `prose_truncated`.
- **Orchestrator wiring.** `reference.md` Step 12.4 calls `render_analysis_json` once
  `analysis.md` is finalized.

## Schema

`candidate_schema.py` is the contract. The `AnalysisReport` envelope holds `report_info`,
`title`, optional `executive_summary`, optional `top_operations`, `compute_optimizations[]`,
and optional `appendix`. A drifted field shows up as a `TypedDict` mismatch at the read
boundary rather than a silent mis-parse. The schema is additive-only, with no
`schema_version` and no fusion or system task shapes, since v1 is compute tier.

## Files

| File | Change |
|---|---|
| `Agent/Analysis/post_processing/analysis_json.py` | new, the renderer (`AnalysisMdParser`, `ComputeGrouper`) |
| `Agent/Analysis/post_processing/candidate_schema.py` | new, the typed `analysis.json` contract |
| `Agent/Analysis/post_processing/__init__.py` | new, public re-exports |
| `Agent/Analysis/utils/validation_utils.py` | `MarkerValidator.OP_ROW_IMPACTS_RE` shared constant |
| `tests/test_analysis_agent_post_processing.py` | new, standalone unit suite (54 tests) |

## Test plan

- [x] Unit suite green: `pytest tests/test_analysis_agent_post_processing.py`, 54 passed.
- [x] **Standalone by construction.** Every fixture is an `analysis.md` baked into the test
  file as a string constant. Nothing is read from `Local_Traces/` or any disk corpus, and
  there is no GPU, LLM, or network. Fixtures use placeholder model names (`ExampleMoE`,
  `ExampleDecoder`) and MI300X's real published peaks (BF16 matrix 708 TFLOPS, HBM 5.3 TB/s,
  from `arch/MI300X.json`), so no proprietary arch fingerprint ships in the open-source
  tests.
- [x] Marker and validator suites are unaffected by the regex de-fork:
  `test_analysis_agent_deterministic.py`, `test_analysis_agent_utils.py`, and
  `test_analysis_agent_evals.py` all pass.
- [x] `black --check` is clean on all changed files, and copyright headers are present.
- [x] **Fresh end-to-end corpus render.** Two inference traces (an MoE model and a decoder
  model) were each run agentic (graph replay plus capture) and deterministic (eager, which
  trips the GRAPH_UNDER_RECORDED fallback). `render_analysis_json` succeeds on all four `analysis.md`.
  On every one the output is schema-valid with no crash, `task.impact.mid` equals the sum of
  the member `impact_score`, `library` is verbatim on agentic and `None` on fallback,
  `analysis_md_rank` is populated, and fan-out rows stay single members. Every compute row
  carried its `op_row` marker (10 of 10, 8 of 8), so real output takes the marker path and
  the heuristic fallback is only exercised by the synthetic fixtures.
- [x] **Impact-fallback coverage.** `test_pct_e2e_fallback_when_no_op_row` (whole block
  absent), `test_null_csv_entry_falls_back_per_row` (per-row `—`), and the parametrized
  `not has_op_row` branch of `test_matrix_impact_path` all assert that `impact_score` equals
  `pct_e2e * HEURISTIC_FRACTION_MID`. A raw `float("—")` on the CSV would have crashed the
  render, and the null now maps cleanly.
- [x] **Determinism regression.** Re-rendering all four fresh reports after the
  heuristic-fallback change produced byte-identical JSON compared to the pre-change output,
  since the change only touches the fallback `else` branch that no real report hits.

## Follow-up (out of scope)

The downstream consumer does not read `analysis.json` yet. It still parses `analysis.md`.
Wiring the consumer to the JSON contract, and retiring the markdown-scraping path, is a
separate change in that repo. Non-compute tiers (fusion, system) and any `schema_version`
are left for a later schema revision.


gabeweisz (approved):
LGTM - nothing controversial here

PR #1048, opened 2026-09-21
Added gsplat 3DGS perf models

Ahmedhasssan-aig:
## Background
3D Gaussian Splatting (3DGS) represents a scene as many 3D Gaussians and renders
by projecting and alpha-blending ("splatting") them to pixels. It gives
photorealistic, real-time novel-view rendering, which is why it's the render
backbone in the 3D world model.

## Summary
The 3DGS kernels are custom CUDA/HIP ops that core TraceLens has no perf model
for, so in the perf report they showed up with empty GFLOPS, data-moved, and
throughput columns. This PR teaches TraceLens how to compute those numbers for
the 3DGS kernels.

## Solution
Add small perf models that estimate the compute (FLOPs) and memory traffic
(bytes) for each 3DGS kernel, reading the tensor shapes from the trace. With
these in place the kernels get proper roofline numbers and show up in a
`GaussianSplat` section of the report.

## Implementation
- New `PerfModel/extensions/gsplat_perf_model_extensions.py` with models for:
  - `_SphericalHarmonics` (spherical-harmonics color eval)
  - `_FullyFusedProjectionPacked` (EWA projection)
  - `_RasterizeToPixels` (per-pixel blending)
- Registered them so TraceLens picks them up automatically: exact-name map in
  `pseudo_ops_perf_utils.py`, plus a substring matcher and category rules in
  `torch_op_mapping.py`.
- Added a model for `aten_upsample_bilinear` (bilinear upsample) that was also
  missing.

## Tests
- New unit tests in `tests/test_gsplat_perf_model.py`; all 24 tests pass
  (including the existing synthetic-op tests).
- Ran a real 3DGS trace end-to-end: the `GaussianSplat` kernels now report
  FLOPs/bytes/throughput and the report's sanity check passes.


TraceLens/PerfModel/extensions/__init__.py
```diff
@@ -61,6 +61,11 @@
     sgl_kernel_qr_all_reduce,
     sgl_kernel_reg_all_gather_into_tensor,
 )
+from .gsplat_perf_model_extensions import (
```
gabeweisz: Discussed offline: follow up PR so we don't need to modify this file or pseudo_ops_perf_utils.py every time we add an extension. The point of extensions is that they are outside the main code and we don't need to update it

gabeweisz (approved):

TraceLens/PerfModel/torch_op_mapping.py
```diff
@@ -64,6 +64,12 @@
 OP_CATEGORY_PATTERNS: List[Tuple[Pattern, str]] = [
     (re.compile(r"^triton"), "triton"),
     (re.compile(r"^record_param_comms"), "record_param_comms"),
+    # gsplat 3DGS kernels; op names embed the symbol behind a launcher/synthetic
+    # prefix, so ".*" is needed (patterns use re.match). projection also gets a
+    # full model via _match_gsplat; the intersect kernels are classify-only
+    # (nnz/n_isects live on the sibling _RasterizeToPixels op).
```
devalshahamd: We can reduce the comments to one line here

TraceLens/PerfModel/torch_op_mapping.py
```diff
@@ -64,6 +64,12 @@
...
+    (re.compile(r".*gsplat::projection_ewa_3dgs_packed_fwd"), "GaussianSplat"),
+    (re.compile(r".*gsplat::intersect_(tile|offset)_kernel"), "GaussianSplat"),
```
devalshahamd: Are these category only ops?

TraceLens/PerfModel/torch_op_mapping.py
```diff
@@ -423,6 +437,25 @@ def _match_triton_compiled(name):
...
+    _RasterizeToPixels / _SphericalHarmonics autograd ops."""
+    from TraceLens.PerfModel.extensions import gsplat_perf_model_extensions as _g
+
+    if "gsplat::projection_ewa_3dgs_packed_fwd" in name:
+        return _g.gsplat_projection_ewa_packed
+    if name.startswith("_RasterizeToPixels"):
+        return _g.gsplat_rasterize_to_pixels
+    if name.startswith("_SphericalHarmonics"):
+        return _g.gsplat_spherical_harmonics
+    return None
```
devalshahamd: Are these category only or perf model based operations. Can we move some mapping to the pseudo_ops_perf_utils.py inside extensions?

TraceLens/PerfModel/torch_op_mapping.py
Ahmedhasssan-aig: Yes, mainly for the GaussianSplat, these are the ops launching the kernels

TraceLens/PerfModel/torch_op_mapping.py
Ahmedhasssan-aig: All three are perf model based and each map to its own perf model class in gsplat_perf_model_extention.py. _SphericalHarmonics and _RasterizeToPixels are already available in pseudo_ops_perf_utils.py. This matcher handles the cases where the exact name is not present, like above ops mostly have some extra suffixes. Therefore, we need substring/prefix matching here.

TraceLens/PerfModel/torch_op_mapping.py
Ahmedhasssan-aig: Sure, will do

PR #1053, opened 2026-09-22
fix: include parent_module in ops_summary merge keys to prevent cross-product

### LINKED ISSUE #1014: compare_perf_reports_pytorch silently produces a cross-product on ops_summary when reports use --group_by_parent_module
## Summary

The `ops_summary` comparison merges on `["name"]` alone. When the input reports are generated with `--group_by_parent_module`, `name` is no longer unique — the report emits one row per `(name, parent_module)` — so the outer merge degenerates into a full cross-product. The extra rows are not marked in any way, and each carries a plausible-looking `_diff` / `_pct` value computed between unrelated modules.

## Repro

Repo at `210fb5c8`:

```bash
TraceLens_generate_perf_report_pytorch_inference \
    --profile_json_path tests/traces/inference/xdit_flux.1/profile_trace_rank_0.json.gz \
    --capture_folder tests/traces/inference/xdit_flux.1/capture_traces \
    --output_xlsx_path base.xlsx --enable_pseudo_ops --group_by_parent_module

TraceLens_generate_perf_report_pytorch_inference \
    --profile_json_path tests/traces/inference/xdit_flux.1_aiter_sage/profile_trace_rank_0.json.gz \
    --capture_folder tests/traces/inference/xdit_flux.1_aiter_sage/capture_traces \
    --output_xlsx_path sage.xlsx --enable_pseudo_ops --group_by_parent_module

TraceLens_compare_perf_reports_pytorch base.xlsx sage.xlsx \
    --names base sage --sheets gpu_timeline ops_summary ops_all roofline \
    -o comparison.xlsx --output_csvs_dir compare_csvs
```

```
base    ops_summary: 101 rows,  68 unique names
sage    ops_summary: 110 rows,  77 unique names
compare ops_summary: 327 rows   <- exactly 101 x 110 cross-product
```

Per-op blowup:

| op | base rows | sage rows | comparison rows |
|---|---|---|---|
| `aten::add` | 10 | 10 | 100 |
| `aten::copy_` | 7 | 7 | 49 |
| `aten::mul` | 6 | 6 | 36 |
| `aten::mm` | 2 | 2 | 4 |

Concretely, `aten::mm` yields four rows, including base `xFuserFlux1Transformer2DWrapper` (202.7 ms) joined against sage `T5Attention/Linear` (12.6 ms), reported as `total_direct_kernel_time_ms__sage_pct = -93.78`. Nothing distinguishes it from the legitimate pairing.

Without `--group_by_parent_module` the same report has 68 rows / 68 unique names and the join is sound.

## Cause

`TraceLens/Reporting/compare_perf_reports_pytorch.py:65-70`:

```python
"ops_summary": {
    "keys": ["name"],
    ...
```

`parent_module` is present as a column in both reports but is not part of the key. The `ops_all` / `ops_unique_args` / `unified_perf_summary` configs key on `name` plus the arg columns and are unaffected.

## Impact

All committed inference reference reports are generated with `group_by_parent_module=True` (see `tests/test_inference_perf_report.py`), so this is the default shape for inference report comparison.

## Suggested fix

Include `parent_module` in `merge_keys` for `ops_summary` when the column is present in every input report. Failing that, detect duplicate keys before merging and raise, rather than emitting silent cross-product rows.

kyle-hoffmeyer:
### Summary

This PR fixes a bug in `compare_perf_reports_pytorch`. `compare_perf_reports_pytorch` merged `ops_summary` sheets on `["name"]` alone. When reports are generated with `--group_by_parent_module` or `is_recompute`, `name` is no longer unique since the report emits one row per `(name, parent_module, is_recompute)` combination, so the outer merge degenerates into a full cross-product.

The fix promotes these optional grouping columns into merge keys when they are present in all input reports, and raises an error when some reports have them and others don't.

### Implementation

**Root cause.** `SHEETS_COMPARE_CONFIG` defined `ops_summary` merge keys as `["name"]` and did not consider the optional grouping columns:

```python
"ops_summary": {
    "keys": ["name"],
    ...
}
```

There are exactly two columns that can make `name` non-unique in `ops_summary`:

| Column | Trigger |
|---|---|
| `parent_module` | `--group_by_parent_module` flag |
| `is_recompute` |  `--detect_recompute` flag |

In `process_summary_sheet`, after loading the dataframes, optional keys are promoted to merge keys when present in all input reports. When a key is present in some reports but not all (i.e. comparing a `--group_by_parent_module` report against a non-grouped one), a `ValueError` is raised, since the granularities are incompatible and merging would still produce a partial cross-product.

### Tests

Three unit tests in `test_compare_perf_reports.py` using synthetic `ops_summary` CSVs via `tmp_path`:

| Test | Setup | Assertion |
|---|---|---|
| `test_ops_summary_no_cross_product_with_parent_module` | Both reports have `parent_module`, 2 rows each with duplicate `name` but distinct `parent_module` | Result has 2 rows (not 4), `parent_module` in output columns |
| `test_ops_summary_merge_without_parent_module` | Neither report has `parent_module`, 2 rows each with unique `name` | Result has 2 rows, `parent_module` not in output columns |
| `test_ops_summary_mismatched_parent_module_raises` | One report has `parent_module`, the other doesn't | `ValueError` raised matching `"parent_module"` |

All three tests verified to fail on the original code and pass on the fix.

Closes #1014


tests/test_compare_perf_reports.py
```diff
@@ -309,3 +312,95 @@ def test_compare_report_matches_reference(generated_reports, update_references):
...
     errors = _validate_report_against_reference(compare_dir, E2E_REF_COMPARE_CSVS)
     assert not errors, "Compare report differences:\n" + "\n".join(errors)
+
+
+# ─── Unit tests for ops_summary cross-product fix (#1014) ────────────────────
+
+
+def _make_ops_summary_csv(path, rows, has_parent_module=True):
+    """Write a minimal ops_summary.csv for testing."""
+    os.makedirs(path, exist_ok=True)
```
gabeweisz: Don't we have a utility class to create CSV files?

gabeweisz: Other than that this looks good to me

kyle-hoffmeyer: The other tests seem to inline this code. There isn't really a general function/class for creating CSV files.
Just for cleanliness, it'd be useful to organize the tests better in another PR though.

